PR #35, opened 2019-06-18
Upgrade dependencies ⏫

lluia:
## Notes 💬

Upgrading dependencies on this project to their latest  ⏫ **minor version** ⏫

### DIFF
--- a/src/setupTests.js
+++ b/src/setupTests.js
@@ -1,6 +1,6 @@
 import 'jest-styled-components';
 import 'jest-dom/extend-expect';
 import 'jest-axe/extend-expect';
-// react-testing-library renders your components to document.body,
+// @testing-library/react renders your components to document.body,
 // this will ensure they're removed after each test.
-import 'react-testing-library/cleanup-after-each';
+import '@testing-library/react/cleanup-after-each';
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "release": "semantic-release -e ./etc/semantic-release/config.js --no-ci"
   },
   "dependencies": {
-    "core-js": "2",
+    "core-js": "^3.1.4",
     "downshift": "^2.2.2",
     "react-modal": "^3.8.1",
     "styled-normalize": "^8.0.6"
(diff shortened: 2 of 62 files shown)


lluia:
@thegrinder @poteirard 

I fixed all the small issues that arose from this update. 

I played around with every component in **Styleguidst** and all seems fine 🤞🏻💃🏼

There's one test still broken: for [`useAccordion`](https://github.com/zopaUK/react-components/blob/master/src/components/hooks/useAccordion/useAccordion.test.ts#L26-L44).

I did a bit of debugging, and for what I can understand, it seems it is a problem of a state value cached in a closure on [this util we expose](https://github.com/zopaUK/react-components/blob/master/src/components/hooks/useAccordion/useAccordion.ts#L28). Somehow is reading the old state rather than the new one [updated here](https://github.com/zopaUK/react-components/blob/master/src/components/hooks/useAccordion/useAccordion.ts#L49-L52).

@thegrinder as you wrote this util, would you mind trying to investigate this?  👀 🙏

lluia:
@thegrinder @poteirard I found [the solution](https://medium.com/@divyabiyani26/react-hooks-with-closures-usestate-v-s-usereducer-9e0c20e81051) for the problem above ☝🏼 , this PR is ready now... ☕️

lluia:
Speaking with @poteirard, we need to make sure the `core-js` upgrade doesn't break this library in **IE11** 🦄

lluia:
@poteirard I just tested on **IE 11** ( _browserstack_ ) and our live site is a blank page 😐

Screenshot 2019-07-14 at 17 18 10

I think is something related to `styled-components` and how to polyfill or minify the source code 🧠

I'm investigating [this issue](https://github.com/styled-components/styled-components/issues/2190), the error given in the console points to **Error 1** on this [reference error page](https://github.com/styled-components/styled-components/blob/master/src/utils/errors.md).

lluia:
Actually, me being silly 🤦🏻‍♂️

As you can see in the URL of the screenshot abover, the current version of this library in `master` already doesn't work on **IE 11** 💃🏼

src/setupTests.js
```diff
@@ -1,6 +1,6 @@
 import 'jest-styled-components';
 import 'jest-dom/extend-expect';
```
thegrinder: I believe jest-dom was also moved if I'm not mistaken

lluia: It seems we still need that otherwise some extensions of `expect` we're using won't be there:
```
TypeError: expect(...).toHaveFocus is not a function
```

thegrinder: that's weird cause it looks like it's there https://github.com/testing-library/jest-dom

Issue #36, opened 2019-06-20
<Modal /> z-index too low


steedems:
The Modal [z-index](https://github.com/zopaUK/react-components/blob/ae12f12e2f33456c6d693cc780e2c6bf1fc44124/src/components/molecules/Modal/ModalStyles/ModalStyles.tsx#L12) of `ModalStyles` should probably higher, or the implementation should allow to easily overwrite it. cc @thegrinder 

![Screenshot 2019-06-20 at 15 51 56](https://user-images.githubusercontent.com/7896422/59854714-b6dd7680-9373-11e9-8c68-2ac234b19106.png)

lluia:
A simple API for this would be to _just_ add a `layer` prop:
```jsx
<Modal layer={100}>
  <Other />
</Modal>
```

PR #40, opened 2019-07-01
Control applied `z-index` in `<Modal />` 👔

lluia:
Addresses #36. 

Adding a `layer` prop to `<Modal.Styles />` to control the `z-index` applied on the rendered modal,
for example:
```jsx
<NavBar /> // this is fixed and internally applies a `z-index` value of `100`
<Modal.Styles layer={200} /> // ensures `<Modal />` will always render on top of `<NavBar />`  
```

### DIFF
--- a/src/components/molecules/Modal/Modal.test.tsx
+++ b/src/components/molecules/Modal/Modal.test.tsx
@@ -3,10 +3,39 @@ import { render, waitForElement } from 'react-testing-library';
 import Modal from './Modal';
 
 describe('<Modal />', () => {
-  it('render the opened Modal', async () => {
+  it('can render un-opened', async () => {
+    Modal.setAppElement('body');
+    const { queryByText } = render(<Modal isOpen={false}>text</Modal>);
+
+    expect(queryByText('text')).toBe(null);
+  });
+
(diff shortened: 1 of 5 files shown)


src/components/molecules/Modal/Modal.test.tsx
```diff
@@ -3,10 +3,22 @@ import { render, waitForElement } from 'react-testing-library';
...
+
+  it('can render opened', async () => {
     Modal.setAppElement('body');
     const { getByText } = render(<Modal isOpen>text</Modal>);
+
     const modal = await waitForElement(() => getByText('text'));
     expect(modal).toMatchSnapshot();
   });
+
+  it('allows to set global styles', () => {
```
lluia: I had to a some more tests as **codecov** wouldn't allow to merge this due to loosing coverage...

It seems [testing the ouptut of `createGlobalStyles`](https://spectrum.chat/styled-components/jest/testing-a-createglobalstyle~7ed712f0-e0f1-430b-85c9-bcc9aa23e077) is no really straightforward either 😐

lluia:
Added some tests covering the **global styles** applied through `<Modal.Styles />`

PR #42, opened 2019-07-12
fix(accordion): add support for different header sizes

JuliaHempel:
Add support for different sized headers on the accordion component. 
https://marvelapp.com/9hj9j4b/screen/48163664

### DIFF
--- a/src/components/organisms/Accordion/Accordion.md
+++ b/src/components/organisms/Accordion/Accordion.md
@@ -11,26 +11,28 @@ const AccordionExample = () => {
   const items = [
     {
       id: 'one',
-      header: 'header one',
+      header: 'header one - default (lead)',
       section: 'section one',
     },
     {
       id: 'two',
-      header: 'header two',
+      header: 'header two - lead',
       section: 'section two',
+      size: 'lead',
     },
     {
       id: 'three',
-      header: 'header three',
+      header: 'header three - regular',
       section: 'section three',
+      size: 'regular',
     },
   ];
 
   return (
     <div aria-label="accordion example">
-      {items.map(({ id, header, section }, index) => (
+      {items.map(({ id, header, section, size }, index) => (
         <div key={id}>
-          <Accordion.Header isOpen={isActiveSection(index)} {...getHeaderProps(id, index)}>
+          <Accordion.Header isOpen={isActiveSection(index)} {...getHeaderProps(id, index)} textSize={size}>
             {header}
           </Accordion.Header>
           <Accordion.Section {...getSectionProps(id, index)}>{section}</Accordion.Section>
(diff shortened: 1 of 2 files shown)


src/components/organisms/Accordion/Accordion.md
```diff
@@ -11,26 +11,28 @@ const AccordionExample = () => {
...
   ];
 
   return (
     <div aria-label="accordion example">
-      {items.map(({ id, header, section }, index) => (
+      {items.map(({ id, header, section, size }, index) => (
         <div key={id}>
-          <Accordion.Header isOpen={isActiveSection(index)} {...getHeaderProps(id, index)}>
+          <Accordion.Header isOpen={isActiveSection(index)} {...getHeaderProps(id, index)} textSize={size}>
             {header}
```
poteirard: I think on of the advantages of using render props is that you can pass whatever component in the render, so you can customise  like this:
```<Lead>{header}</Leader>```
and then you don't need custom sizes that can lead to more changes in the future. /cc @thegrinder

File: src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx [hunk 2]
```diff
@@ -34,14 +44,15 @@ const Title = styled.span`
 export interface IAccordionHeader {
   /** determines the position of the chevron icon */
   isOpen: boolean;
+  textSize?: TTextSize;
 }
 
 const AccordionHeader: FunctionComponent<IAccordionHeader> = React.forwardRef<HTMLButtonElement, IAccordionHeader>(
-  ({ children, isOpen, ...rest }, ref) => (
+  ({ children, isOpen, textSize = 'lead', ...rest }, ref) => (
```
thegrinder: it's required from the looks of it, so not sure if you need the default here

thegrinder: probably no need for a default here, since it's required

src/components/organisms/Accordion/Accordion.md
thegrinder: I guess if Accordion Header has a specific set of sizes, it makes sense to add them as a prop. We just have to make sure we limit the amount of variations

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
thegrinder: no need for default here or we can make it optional?

JuliaHempel: I didn't mean to make it a required prop, it should be optional.

src/components/organisms/Accordion/Accordion.md
JuliaHempel: I agree with @thegrinder here. We may not want too many different variations on the accordion so limiting it by adding the prop seems like a good way for me. It will give us a default way that you can overwrite like @poteirard says here.

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
thegrinder: my bad, it's all good. I looked at the text props by mistake

PR #43, opened 2019-07-15
fix(text input): add support for prefixes

JuliaHempel:


src/components/molecules/TextField/TextField.tsx
```diff
@@ -5,28 +5,62 @@ import InputLabel from '../../atoms/InputLabel/InputLabel';
...
+}
+
+const Prefix = styled.span<IPrefixProps>`
+  position: relative;
+  
+  &::before {
+    content: '${({ prefix }: IPrefixProps) => prefix}';
+    position: absolute;
+    top: -4px;
+    left: 10px;
```
thegrinder: it would be more flexible if you do it this way.
```
position: absolute;
top: 0;
bottom: 0;
display: flex;
align-items: center;
...
```
and give it display block to the `span`

src/components/molecules/TextField/TextField.tsx
```diff
@@ -5,28 +5,62 @@ import InputLabel from '../../atoms/InputLabel/InputLabel';
...
   );
 };
 
+interface IPrefixProps extends HTMLSpanElement {
+  prefix: string;
+}
+
+const Prefix = styled.span<IPrefixProps>`
+  position: relative;
+
```
thegrinder: as mentioned below, `display: block;` here

PR #48, opened 2019-07-31
fix(text field): add ability to style the TextField

JuliaHempel:
Changed the fraction wrapper to be a div so we can add styling when using the TextField.

### DIFF
--- a/src/components/molecules/TextField/TextField.tsx
+++ b/src/components/molecules/TextField/TextField.tsx
@@ -37,14 +37,14 @@ const TextField = (props: IField) => {
     />
   );
   return (
-    <>
+    <div {...rest}>
       {label && <InputLabel htmlFor={`text-id-${name}`}>{label}</InputLabel>}
       {helpText && <HelpText>{helpText}</HelpText>}
       <SizedContainer size={size} {...rest}>
         {prefix ? <Prefix prefix={prefix}>{input}</Prefix> : input}
       </SizedContainer>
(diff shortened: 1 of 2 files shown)


src/components/molecules/TextField/TextField.tsx
```diff
@@ -37,14 +37,14 @@ const TextField = (props: IField) => {
     />
   );
   return (
-    <>
+    <div {...rest}>
```
thegrinder: not sure if need that, you can just do that outside of the component. Also, notice that we're already spreading props over `<SizedContainer>`

lluia: @JuliaHempel I think as @thegrinder ☘️ says you can do instead:
```js
const CustomField = styled(TextField)`
  /* any CSS you see fit here */
`
```
If you need to styled the input specifically:
```js
const CustomInput = styled(InpuText)`
  /* any CSS you see fit here */
`
```

thegrinder: I was thinking more of something like that:
```
const Wrapper = styled.div`
  /* styles */
`
<Wrapper>
  <TextField />
</Wrapper>
```
if you use `styled()`, it's gonna overwrite styles of just the input wrapper, not a wrapper around the input, label, and error message. It really depends on the use case

JuliaHempel: You're right. I'll just wrap it in order to style it and close this.

JuliaHempel:
Not needed

Issue #30, opened 2019-06-04
Remove social media icons 


lluia:
I noticed we have a couple of components ( `<Facebook />` and `<Twitter />`) that render social media icons for **Twitter** and **Facebook**:

<img width="129" alt="Screenshot 2019-06-04 at 17 56 19" src="https://user-images.githubusercontent.com/5938217/58898471-f3b23800-86fa-11e9-8bdc-9fef71c67b40.png">

<img width="131" alt="Screenshot 2019-06-04 at 17 58 13" src="https://user-images.githubusercontent.com/5938217/58898472-f4e36500-86fa-11e9-947e-3ad599d364c6.png">

They're just used within `<ZopaFooter />` and we could _depecrate_ and remove them in upcoming major version?

I can't see the benefit of exporting those to build other stuff 🤔

poteirard:
Related https://github.com/zopaUK/react-components/issues/15

PR #55, opened 2019-08-02
Don't export Facebook and Twitter icons 🦅

lluia:
## Summary

We were exporting two icons related to social media platforms:

- [`<Twitter />`](https://github.com/zopaUK/react-components/blob/master/src/components/icons/Twitter/Twitter.tsx)
- [`<Facebook />`](https://github.com/zopaUK/react-components/blob/master/src/components/icons/Facebook/Facebook.tsx)

These icons were only used within our `<ZopaFooter />`.

This PR removes these exports and inlines the icons instead.

It's marked as a **breaking change**. 

## Issue
- #30


src/components/molecules/ZopaFooter/__snapshots__/ZopaFooter.test.tsx.snap
```diff
@@ -597,15 +597,15 @@ and we are registered with the Office of the Information Commissioner (No. Z8797
           >
             <svg
               fill="currentColor"
-              height="30px"
+              height="24px"
```
thegrinder: is the size change intended here?

lluia: Good catch 🧐 💯 

I see [on `<SocialLinks />`](https://github.com/zopaUK/react-components/blob/master/src/components/molecules/ZopaFooter/SocialLinks/SocialLinks.tsx#L33) we were explicitly setting their width at `30px` however their default one was set to `24xpx` 🤷🏻‍♂️

I'll update so that both have a with of `30px` instead. 👍🏻

PR #60, opened 2019-08-05
fix(footer): fix car loans link and copyright year

rominamoya:
Changed the car finance link  to car loans and changed to dynamic copyright year in the legal text to show actual year


src/components/molecules/ZopaFooter/__snapshots__/ZopaFooter.test.tsx.snap
```diff
@@ -82,7 +82,7 @@ and entered on the Financial Services Register under firm registration number 71
       <p
         class="c3"
       >
-        © Zopa Limited 2018 All rights reserved. 'Zopa' and the Zopa logo are trade marks of Zopa Limited. 
+        © Zopa Limited 2019 All rights reserved. 'Zopa' and the Zopa logo are trade marks of Zopa Limited.
```
poteirard: If this year is dynamic we will need to update the snapshot manually every year. Unless we mock the date which I think it's the best approach in this case. If you need help let me know :)

rominamoya: @poteirard agree, and done!

PR #62, opened 2019-08-07
Use form

thegrinder:
First version of the hook to manage forms


src/components/hooks/useForm/useForm.ts
```diff
@@ -0,0 +1,74 @@
+import { useState, useCallback, useMemo } from 'react';
+
+interface TDictionary<T> {
```
poteirard: I think you can just use `Record` for that

src/components/hooks/useForm/useForm.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  afterEach(() => {
+    onSubmit.mockClear();
+  });
+
+  it('should return the correct object', () => {
+    const { result } = renderHook(() =>
+      useForm({
+        initialValues,
+        onSubmit,
+        validate,
```
poteirard: there's a type error here:
```
Error:(36, 9) TS2322: Type '(values: TTestForm) => Partial<TTestForm>' is not assignable to type 'TValidate'.
  Types of parameters 'values' and 'values' are incompatible.
    Type 'TDictionary<any>' is missing the following properties from type 'TTestForm': name, lastName
```

thegrinder: modified types and that shouldn't happen anymore. Also, I'm gonna be looking at improving them

src/components/hooks/useForm/useForm.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      }),
+    );
+    ['getFieldProps', 'invalid', 'submitting', 'handleSubmit'].forEach(property => {
```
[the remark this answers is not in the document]
thegrinder: you're right, I'll remove the snapshot for both

src/components/hooks/useForm/useForm.ts
```diff
@@ -0,0 +1,68 @@
...
+        onSubmit(values);
+      }
+    },
+    [invalid, onSubmit, values],
+  );
+
+  return {
+    getFieldProps,
+    invalid,
+    submitting,
```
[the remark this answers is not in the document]
thegrinder: I'm probably getting ahead of myself here - thinking about manually feeding context to form provider. I'll remove that and we can think about it again once we need it

thegrinder:
@lluia I'm gonna be building form components around this hook soon and the tests there are gonna be written in a way that you show above. I think we're aiming to have two form options: hook-only solution and set of components that use that hook internally

lluia:
@thegrinder coolios, happy to leave tests as they're now then 💪🏻

PR #76, opened 2019-08-27
fix(rollup): upgrade config and deps

poteirard:
Rollups has been updated
Terser plugin has been added to minify the compiled files
Add browserlist config as shopfront has
Remove unused config (it's already in babel-react-app-preset)


src/components/molecules/TextField/TextField.tsx
```diff
@@ -6,7 +6,7 @@ import InputText from '../../atoms/InputText/InputText';
 import SizedContainer from '../../layout/SizedContainer/SizedContainer';
 import { IField } from '../../types';
 import styled from 'styled-components';
-import { colors } from '../../..';
```
poteirard: fix circular dependency

PR #80, opened 2019-08-27
Improve accordion

thegrinder:
Improve accordion organims
- pass accordion state/props implicitly through context
- Documenting both, <Accordion /> component and the entire organism


src/components/hooks/useAccordion/useAccordion.ts
```diff
@@ -2,6 +2,40 @@ import { useRef, useState, Dispatch, SetStateAction } from 'react';
 import { isArrowDown, isArrowUp } from '../../../helpers/keyboard-keys';
 import { mod } from '../../../helpers/utils';
 
+interface IAccordionHeaderProps {
+  'aria-controls': string;
```
poteirard: is there a reason why we are not extending the Button props instead of adding manually every prop?

poteirard: Same in line 25 for `div`

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.md
```diff
@@ -12,13 +12,13 @@
 ```js
 import { Accordion } from '@zopauk/react-components';
 
-<Accordion.Header isOpen={true}>opened accordion header</Accordion.Header>;
+// <Accordion.Header isOpen={true}>opened accordion header</Accordion.Header>;
```
poteirard: is this intended?

src/components/hooks/useAccordion/useAccordion.ts
thegrinder: this way they are required and not optional which makes your life easier when adding custom input components

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.md
thegrinder: nope, was work in progress before, fixed now

endymion1818:
If JS is disabled for a user, does this render the accordion in an open state?

thegrinder:
@endymion1818 I'm not sure how relevant this edge case is in the world where everything is probably broken  without JS. This PR is concerned only with refactoring accordion to use react context api. However, if you feel that it's something that needs to be addressed, feel free to open a PR.

poteirard:
As @endymion1818 said if JS is disabled none of the products in Zopa will work as they are SPA. I just checked zopa.com and while it works on the home it doesn't work on https://www.zopa.com/loans anyway

endymion1818:
Yes I think it's out of scope for any of the zopa apps, we're hoping that eventually all critical journeys through Shopfront will have that feature though, mostly for SEO implications but also for users.

src/components/hooks/useAccordion/useAccordion.ts
```diff
@@ -2,6 +2,40 @@ import { useRef, useState, Dispatch, SetStateAction } from 'react';
...
+  'aria-disabled': boolean;
+  'aria-expanded': boolean;
+  id: string;
+  key: string;
+  onClick: () => void;
+  onFocus: () => void;
+  onKeyDown: (e: React.KeyboardEvent<HTMLButtonElement>) => void;
```
poteirard: interesting... React is not imported but still works 🤔

thegrinder: no jsx :)

poteirard: But still `React.KeyboardEvent` is not imported but I guess because React refers to the namespace in the type definitions instead of the React object itself? 🤔

thegrinder: oh yeah, you're right... hmmm 🤔

thegrinder: imported it, just in case

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -29,34 +31,34 @@ const Title = styled(Text)`
 `;
 
 export interface IAccordionHeader {
-  /** determines the position of the chevron icon */
-  isOpen: boolean;
+  id: string;
+  index: number;
   textSize: ITextProps['size'];
```
lluia: It looks like `textSize` is defaulting to `"medium"` so there's no need to make it required?
 ( _I know it was like this already_ 🙈  )

src/components/organisms/Accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import Accordion from '..';
+
+describe('<Accordion />', () => {
+  it('should render aria-label', () => {
+    const ariaLabel = 'test-accordion';
+    const { queryByText } = render(<Accordion aria-label={ariaLabel} />);
+    expect(queryByText).toBeTruthy();
```
lluia: Hmm, this doesn't seem right? 🤔 
( `queryByText` is always _truthy_ as it's a function 😅 )

src/components/hooks/useAccordion/useAccordion.ts
```diff
@@ -1,14 +1,48 @@
-import { useRef, useState, Dispatch, SetStateAction } from 'react';
+import { useRef, useState, Dispatch, SetStateAction, KeyboardEvent, RefObject } from 'react';
 import { isArrowDown, isArrowUp } from '../../../helpers/keyboard-keys';
 import { mod } from '../../../helpers/utils';
 
+interface IAccordionHeaderProps {
+  'aria-controls': string;
+  'aria-disabled': boolean;
+  'aria-expanded': boolean;
+  id: string;
```
lluia: If we end-up setting `id` on the HTML elements rendered by `<Accordion.Header />` and `<Accordion.Section />` there's a chance that we end with two duplicate `id`s on the document ?
```jsx
<Accordion aria-label="test-accordion">
  <Accordion.Header id="one" />
  <Accordion.Section id="one" />
</Accordion>;
```
I believe that's bad as `id` should be unique? Maybe we could use `data-accordion-id` instead?

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -29,34 +31,34 @@ const Title = styled(Text)`
...
+  large: '14px',
+  medium: '10px',
+  small: '8px',
+};
+
+const AccordionHeader: FC<IAccordionHeader> = ({ children, id, index, textSize = 'medium', ...rest }) => {
+  const { getHeaderProps, isActiveSection } = useAccordionContext();
+  const { ref, ...headerPropsRest } = getHeaderProps(id, index);
+  return (
+    <StyledButton ref={ref} {...headerPropsRest} {...rest}>
```
lluia: I believe `<StyledButton />` doesn't need the `ref`, same for 
```js
<div ref={ref}>
``` 
in `<Accordion.Section />` ? 🤔

src/components/hooks/useAccordion/useAccordion.ts
thegrinder: this `id` doesn't translate directly to the `id` attribute in case. The above example would result in:
```
<Accordion.Header id="one" />
<Accordion.Section id="one-section" />
```

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
thegrinder: it needs it to properly handle focus

PR #108, opened 2019-09-23
feat(trustpilotlogo): add trustpilot logo component

endymion1818:
Add Trustpilot logo to the component library


src/components/icons/TrustPilotLogo/TrustPilotLogo.md
```diff
@@ -0,0 +1,7 @@
+This component is a wrapper around the svg code for Zopa's logo.
```
poteirard: 🤔

endymion1818: 😆  whoops

lluia:
@endymion1818 

I would prefer to keep logos ( _apart from Zopa's one_ ) out of the component library 💅🏻 , I believe this logo is also quite specific to _Shopfront_ 🛍 ?

The few current icons we expose might even be removed in the future as we plan to use the same icon font the **mobile app** is using.

endymion1818:
@lluia 

https://github.dns.ad.zopa.com/search?q=trustpilotlogo&type=Code

Auto-comparison and ourselves are the only ones using it but it could become more prominent further down the brand refresh route.

I asked Alex TerHark about this, he says ideally this and another icon (the trustpilot stars) should be in here, there is a risk we'll update them when we get to the stage where icons are being considered separate from use (borrowing power page) however it would be preferable for them to be added now if possible ... 

You didn't just say _icon fonts_ did you? 😱 https://twitter.com/sarasoueidan/status/639013751644098561

endymion1818:
Closing after a discussion about which components would be appropriate for this library.

PR #112, opened 2019-09-25
fix(peer dependencies): move styled components to peer dependencies

JuliaHempel:
update it to be a version after 4.3 since that's when forwardedAs was introduced.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "peerDependencies": {
     "react": "^16.8.1",
-    "styled-components": "^4.0.0",
+    "styled-components": "^4.4.0",
     "typescript": ">=3.x <4"
   },
   "devDependencies": {
@@ -85,7 +85,6 @@
     "rollup-plugin-terser": "^5.1.1",
     "rollup-plugin-url": "^2.2.2",
     "semantic-release": "^15.13.12",
-    "styled-components": "^4.1.3",
     "typescript": "^3.2.2"
   },
   "browserslist": [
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -85,7 +85,6 @@
     "rollup-plugin-terser": "^5.1.1",
     "rollup-plugin-url": "^2.2.2",
     "semantic-release": "^15.13.12",
-    "styled-components": "^4.1.3",
     "typescript": "^3.2.2"
```
lluia: I think we still need `styled-components` declared in `devDependencies` so that when running the project locally to develop new components it can find  `styled-components` in `node_modules` 👍

src/components/atoms/Badge/__snapshots__/Badge.test.tsx.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`<Badge /> renders with no a11y violations 1`] = `
-.c1 {
```
lluia: I guess all these snapshot changes are from bumping `styled-components` ? 🤔

lluia:
Yup, we'll need to put `styled-components` back on `devDependencies` 👍 , also let's run `yarn` after that to make sure `yarn.lock` gets updated 🗝

src/components/atoms/Badge/__snapshots__/Badge.test.tsx.snap
lluia: Yup tried locally and can confirm it comes from bumping `styled-components` 👍

PR #118, opened 2019-10-03
fix(footer colours): update footer to use Zopa colours

JuliaHempel:
Update the footer to only use the new colours.

### DIFF
--- a/src/components/molecules/ZopaFooter/FooterLink/FooterLink.tsx
+++ b/src/components/molecules/ZopaFooter/FooterLink/FooterLink.tsx
@@ -1,17 +1,17 @@
 import styled from 'styled-components';
-import * as colors from '../colors';
 import Link from '../../../atoms/Link/Link';
+import { colors } from '../../../../constants/colors';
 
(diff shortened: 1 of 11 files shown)


lluia:
Thanks for the suggestion @JuliaHempel ❤️ 🎉 

We're aware that `<ZopaFooter />` was the only component that is using [a specific set of colors](https://github.com/zopaUK/react-components/blob/master/src/components/molecules/ZopaFooter/colors.ts).

The reason for this being that the neutral colors in [our new colors scheme](https://zopauk.github.io/react-components/#/Content?id=colors) didn't account for all the grey variations that existed on this component:



Making it match our new colors creates **two side effects**:



- everything becomes "quite white" so the links are not easily differentiated
- the `:hover` state makes the text quite dark and hard to read

I don't mind doing the update, but lets check what other teams think about it 👍 
@poteirard  @AntonyKwok @endymion1818

endymion1818:
Yes! Nice catch @lluia @JuliaHempel

JuliaHempel:
replaced by: https://github.com/zopaUK/react-components/pull/119

PR #120, opened 2019-10-11
refactor(input-label): change input-label for label

### LINKED ISSUE #49: Rename `<InputLabel />` to `<Label />`
Our `<InputLabel />` atom is just [a styled HTML `label`](https://github.com/zopaUK/react-components/blob/master/src/components/atoms/InputLabel/InputLabel.tsx#L7-L14).

It can be used to map both from inputs or other elements like `select` and `textarea`, hence it would be less confusing if we just name it as `<Label />`. 

Otherwise might think from the name that's only meant to be used with `<InputText />`.

dfvalero:
BREAKING CHANGE: input-label won't be available anymore

fix #49


lluia:
Awesome @dfvalero 💯 

Code looks good 💅🏻 👍🏻

Could we change the base branch to be `milestone/3.0`  instead of `master`? 

We usually pack breaking changes on big releases, we just [entered `2.x.x`](https://github.com/zopaUK/react-components/milestone/1) and probably we're gonna wait a couple of months before we go into `3.x.x`.

Teams are still migrating to `2.x.x` and we want to wait a bit until we introduce more breaking changes 💆🏻‍♂️

dfvalero:
> Awesome @dfvalero 💯
> 
> Code looks good 💅🏻 👍🏻
> 
> Could we change the base branch to be `milestone/3.0` instead of `master`?
> 
> We usually pack breaking changes on big releases, we just [entered `2.x.x`](https://github.com/zopaUK/react-components/milestone/1) and probably we're gonna wait a couple of months before we go into `3.x.x`.
> 
> Teams are still migrating to `2.x.x` and we want to wait a bit until we introduce more breaking changes 💆🏻‍♂️

done!

PR #124, opened 2019-10-17
Privacy notice

endymion1818:
added privacy notice link back in ... I think I forgot to add it initially!

Added the Question label, just checking with @grabbeh


lluia:
@endymion1818 I think the snapshots updates in all tests might be because you ran a different version of **Jest**. 

Could you revert that? 🙏🏻

PR #127, opened 2019-10-23
Fix types 🌋

lluia:
This brings **two important changes** to the library regarding types:

1. In `package.son` → `devDependencies` ...

2. In `tsconfig.json` ...

We didn't had the types for `styled-components` because they had a long standing issue of adding [`@types/react-native` when installed](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311), which produced all sorts of weird type errors... 🙄

This is being sorted now and we're now able to bring back `--noImplicitAny`, which in turn raised a load of errors I had to fix. 

 It's now ready to be reviewed, I've tried it in **Zopa Savings** and I'm confident the changes are stable enough 🤞🏻


src/components/atoms/Button/Button.tsx
```diff
@@ -99,33 +101,34 @@ const SButton = styled.button<IButtonProps>`
...
+    box-shadow: ${({ styling = 'primary' }) => activeBoxShadows[styling]};
     ${({ styling }) => (styling === 'contrastLink' || styling === 'link' ? null : 'outline: none')};
   }
 
   &:hover:not(:disabled) {
     opacity: 0.8;
-    color: ${({ styling }) => hoverFontColors[styling]};
-    background-color: ${({ styling }) => activeBackgroundColors[styling]};
+    color: ${({ styling = 'primary' }) => hoverFontColors[styling]};
+    background-color: ${({ styling = 'primary' }) => activeBackgroundColors[styling]};
```
poteirard: I think it will look cleaner to have
```ts
SButton.defaultProps = {
  styling: 'primary',
  sizing: 'default',
}
```
So we don't have to repeat that login in the props func all the time

src/components/atoms/Link/Link.tsx
```diff
@@ -1,15 +1,20 @@
...
 import styled from 'styled-components';
-import { colors, IBaseColorSpec, INeutralColorSpec } from '../../../constants/colors';
+import { colors } from '../../../constants/colors';
 import Text from '../../atoms/Text/Text';
 
-export interface ILinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
-  color?: IBaseColorSpec['secondary'] | INeutralColorSpec['white'];
+export interface ILinkProps
+  extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
+    React.RefAttributes<HTMLAnchorElement> {
```
poteirard: You can use this instead of those 2 extensions: `React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>`
Actually I think we should be using always this. As this is the return type of `styled.a`

src/components/atoms/Link/Link.tsx
```diff
@@ -42,12 +47,13 @@ const TargetIcon = (props: ITargetIconProps) => {
   );
 };
 
-const Link: FC<ILinkProps> = React.forwardRef<HTMLAnchorElement, ILinkProps>((props, ref) => {
-  const { children, color = colors.base.secondary, ...rest } = props;
+const Link = React.forwardRef<HTMLAnchorElement, ILinkProps>((props, ref) => {
+  const { children, color = colors.base.secondary, target, ...rest } = props;
```
poteirard: this line is not needed:
```
const Link = React.forwardRef<HTMLAnchorElement, ILinkProps>(({ children, color = colors.base.secondary, target, ...rest }, ref) => {
```

src/components/hooks/useForm/useForm.ts
```diff
@@ -16,8 +14,8 @@ export interface IFieldProps {
   error: string | undefined;
   touched: boolean;
   value: TValue;
-  onChange: (e: any) => void;
-  onBlur: () => void;
+  onChange: (eventData: unknown) => void;
+  onBlur: (eventData: unknown) => void;
```
poteirard: maybe a better type would be:
```
  onChange: React.ChangeEventHandler<Element>;
  onBlur: React.FocusEventHandler<Element>;
```

src/components/icons/Chevron/Chevron.tsx
```diff
@@ -16,14 +16,17 @@ export interface IChevronProps extends React.SVGProps<SVGSVGElement> {
...
     down: '0',
     left: '90',
     right: '-90',
     up: '180',
   };
-  return directionMap[direction] || direction;
+
+  if (typeof direction === 'number') return direction.toString();
+
+  return directionMap[direction];
```
poteirard: I think it looks better like this:
```ts
return typeof direction === 'number' ? direction.toString() : directionMap[direction];`
```
Also I think we should avoid having conditionals in on line but that's another topic

src/components/layout/FlexCol/FlexCol.tsx
```diff
@@ -42,15 +42,19 @@ const genAutoWidth = breakpoint => css`
...
     flex: 0 0 ${(colWidth / cols) * 100}%;
     max-width: ${(colWidth / cols) * 100}%;
+    `
+      : null}
   }
 `;
 
-const getWidth = (breakpoint, colWidth, cols) => {
+const getWidth = (breakpoint: TGridBreakpoints, colWidth: TColWidth, cols: IFlexColProps['cols']) => {
   if (colWidth === 'hidden') {
```
poteirard: I know it's not on this scope but I think this would be better like this:

```ts
const getWidth = (breakpoint: TGridBreakpoints, colWidth: TColWidth, cols: IFlexColProps['cols']) => {
  switch (colWidth) {
    case 'hidden':
      return genHidden(breakpoint);
    case 'auto':
      return genAutoWidth(breakpoint);
    case 'fill':
      return genFillWidth(breakpoint);
    default:
      return genRelativeWidth(breakpoint, colWidth, cols);
  }
};

```

src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
```diff
@@ -2,11 +2,11 @@ import React, { FC } from 'react';
 import styled from 'styled-components';
 import { colors } from '../../../../constants/colors';
 import Chevron from '../../../icons/Chevron/Chevron';
-import Link from '../../../atoms/Link/Link';
+import Link, { ILinkProps } from '../../../atoms/Link/Link';
 
 export interface IStyledNavbarLinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
   active: boolean;
-  color: string;
+  color?: ILinkProps['color'];
```
poteirard: Why not `IStyledNavbarLinkProps extends ILinkProps` ? and then you can get rid of this

src/constants/grid.ts
```diff
@@ -1,4 +1,14 @@
-const grid = {
+export type TGridBreakpoints = 'xs' | 's' | 'm' | 'l' | 'xl';
+export type TGridWidths = 's' | 'm' | 'l' | 'xl';
+
+interface IGrid {
+  breakpoints: { [k in TGridBreakpoints]: number };
```
poteirard: `breakpoints:  Record<TGridBreakpoints, number>;`

src/constants/grid.ts
```diff
@@ -1,4 +1,14 @@
...
+  cols: number;
+  gutter: number;
+  width: { [k in TGridWidths]: number };
```
poteirard: `breakpoints:  Record<TGridWidths, number>;`

src/helpers/responsiveness.ts
```diff
@@ -11,22 +9,24 @@ const sizes = {
...
-    @media (max-width: ${sizes[label]}px) {
-      ${style};
-    }
-  `;
-  return accumulator;
-}, {});
+const maxMedia = {
+  phone: interpolate('phone', 'max'),
+  tablet: interpolate('tablet', 'max'),
+  desktop: interpolate('desktop', 'max'),
```
poteirard: Good refactor :)

src/components/atoms/Button/Button.tsx
lluia: Good point, I think `defaultProps` were previously declared on the wrapper and not on the _styled component_, hence TS wasn't able to infer them. I'll update it 👌

src/components/hooks/useForm/useForm.ts
lluia: It would, however in this implementation `onChange` and `onBlur` not always expect the actual event object, sometimes [they receive the value directly](https://github.com/zopaUK/react-components/blob/master/src/components/organisms/Form/FormRadioField/FormRadioField.tsx#L14) ( cc. @thegrinder  ) hence I typed it as `unknown`.

I'll try to refactor the `<Form.* />` components so that they all send the actual event to this handlers 💆🏻‍♂️

src/components/icons/Chevron/Chevron.tsx
lluia: ha ha, I ❤️ one line conditionals 💅🏻 😛 , can't see nothing wrong with them, seems a bit biased as some people think they don't read good and other like me think they read better 💆🏻‍♂️

src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
lluia: Hmm good point 🤔 , I'll give it a go 👍🏻

src/components/layout/FlexCol/FlexCol.tsx
lluia: Yup don't mind to try to refactor it 👍🏻

src/components/atoms/Link/Link.tsx
lluia: I think we still need to reference `React.RefAttributes` in the type somehow, otherwise TS complains about `ref` below 😪

lluia:
@poteirard addressed your suggestions 💅🏻 👍🏻  apart from the one referring to `<Link />` type extension ( _see my comment_ ), could you have a second look 🧐 ? 🙏🏻

PR #133, opened 2019-10-29
fix(button): add styles for button with as prop

### LINKED ISSUE #129: Secondary Button Border disappears
When defining a secondary styled button as an anchor tag the button loses its border.

This is caused by the `:enabled`

This is a screenshot of the broken button with `as="a"` property set. as you can see the border is missing.

Screenshot 2019-10-28 at 08 19 09

dfvalero:
fix #129

### DIFF
--- a/src/components/atoms/Button/Button.tsx
+++ b/src/components/atoms/Button/Button.tsx
@@ -108,29 +108,28 @@ const SButton = styled.button<IButtonProps>`
   cursor: pointer;
   border: 1px solid ${({ styling }) => backgroundColors[styling]};
   border-radius: 8px;
-  color: ${({ contrastColor }) => contrastColor};
-  color: ${({ styling }) => fontColors[styling]};
+  color: ${({ styling, contrastColor }) => (contrastColor ? contrastColor : fontColors[styling])};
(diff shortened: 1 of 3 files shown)


src/components/atoms/Button/Button.tsx
```diff
@@ -108,29 +108,28 @@ const SButton = styled.button<IButtonProps>`
...
   border-radius: 8px;
-  color: ${({ contrastColor }) => contrastColor};
-  color: ${({ styling }) => fontColors[styling]};
+  color: ${({ styling, contrastColor }) => (contrastColor ? contrastColor : fontColors[styling])};
   background-color: ${({ styling }) => backgroundColors[styling]};
   transition: all 140ms ease-in-out;
 
-  &:enabled {
+  &:not(:disabled) {
     box-shadow: ${({ styling }) => boxShadows[styling] || 'none'};
```
thegrinder: Wouldn't it be easier to just move that box-shadow outside of the enabled? Or am I missing something?

lluia: I tried few months ago to not use actual `border` instead of faking it with a `box-shadow` 🙄 but got tired and left as it was 😪 😅

PR #136, opened 2019-10-30
fix(FlexRow): update types to include missing variants

moshie:
This will add missing types for flex box's `align-items` & `justify-content` properties.


moshie:
@lluia I have updated the commit message I think I misunderstood commitizen's prompt 🙄

PR #137, opened 2019-11-01
fix(prettier): update file name to start with a dot

moshie:


poteirard:
Remove also the `--ignore-path` attribute in the package.json

PR #148, opened 2019-12-09
Treat components as native HTML elements ✨

lluia:
## Summary

This PR aims to fix two issues related to types:

#### 1) Treat and type components as HTML elements

Many times ( _not always_ ) we were spreading the `rest` props on our components in order to enable the client pass native HMTL attributes to them:
```jsx
const Foo = ({ children, ...rest }) => <span {...rest}>{children}</span>
<Foo className="bar" />
```
Now the intend is for every component to allow this behavior and get them typed correctly with `React.HTMLAtrributes` 💆🏻‍♂️

#### 2) Only wrap styled-components when necessary

_Styleguidist_, which uses `react-docgen` internally to [locate components](https://react-styleguidist.js.org/docs/components.html) and generate documentation, wasn't able to locate components when declared just as a `styled-components`:
```js
const Text = styled.span`
  color: pink;
`
export default Text
```
[There's suggested fixes for this](https://react-styleguidist.js.org/docs/thirdparties.html#styled-components) but not supported by `react-docgen-typescript` yet ... 🤯

A temporal solution is to export wrappers as a named export that _Styleguidist_ can locate.

Nudge me in case you want more details on this, the solution is tricky to summarise... 😛

## Try it out

```
$ yarn add @zopauk/react-components@next
$ yarn tsc --noEmit
```

### DIFF
--- a/src/components/atoms/Card/Card.tsx
+++ b/src/components/atoms/Card/Card.tsx
@@ -15,7 +15,9 @@ const borderRadius: { [index in TCardTypes]: string } = {
   linkCard: '8px',
 };
 
-const SCard = styled.div<ICardProps>`
+const Card = styled.div.attrs({
+  as: 'div',
+})<ICardProps>`
   background-color: ${colors.neutral.white};
   border-color: ${colors.neutral.white};
   border-radius: ${({ type }) => borderRadius[type as TCardTypes]};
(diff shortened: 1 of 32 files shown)


src/components/atoms/Card/Card.tsx
```diff
@@ -39,12 +39,13 @@ const SCard = styled.div<ICardProps>`
...
 
-const Card: React.FunctionComponent<ICardProps> = props => {
-  return <SCard {...props} />;
-};
-
 Card.defaultProps = {
   type: 'card',
 };
 
+// TODO: Styleguidist to be able to locate styled components. See #147.
```
lluia: @poteirard I was trying to do the following as per our conversation:
```ts
interface IStyledComponentProps {
  as?: string;
  forwardedAs?: string;
  css?: string;
}

export interface ICardProps extends HTMLAttributes<HTMLDivElement>, IStyledComponentProps {
  type?: TCardTypes;
  display?: 'block' | 'none' | 'inline-block';
}
```
However given we spread props at the bottom **TS** complains about types mismatch:
 
Screenshot 2019-12-10 at 11 17 02

I'm unsure whether there's a straightforward solution to this, any ideas?

PR #164, opened 2020-01-14
New <Tooltip />  component 🧪

lluia:
## Summary

New `<Tooltip />` component which in turns is just a wrapper over [`@tippy.js/react`](https://github.com/atomiks/tippy.js) with design constraints applied on top.

## Bundle cost

**Tippy JS** handles already a lot of the edge-cases associated to tooltips for us ( 👀  **"Adaptability"** section below ) and make sure [they stay as accessible as possible](https://atomiks.github.io/tippyjs/accessibility/#interactivity). 

This comes with an import cost though ☝🏻 ( what do you think @poteirard ? ) but at least we think is safer and more performant than any custom solution we could engineer...  

### 3. Fit on scroll

( &nbsp;👀  _notice how the tooltip vertical alignment changes..._ )

### 4. Fit on viewport edge

( &nbsp;👀 _notice how the tooltip horizontal alignment changes..._ )

### DIFF
--- a/src/components/molecules/Tooltip/Tooltip.test.tsx
+++ b/src/components/molecules/Tooltip/Tooltip.test.tsx
@@ -1,0 +1,18 @@
+import { axe } from 'jest-axe';
+import React from 'react';
+import { render } from '@testing-library/react';
+import Tooltip from './Tooltip';
+
+describe('<Tooltip />', () => {
+  it('renders the component with no a11y violations', async () => {
+    const { container } = render(
+      <Tooltip content="I prefer 🍕 over 🍰 ...">
+        <span>💁🏻‍♂️</span>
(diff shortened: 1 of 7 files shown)


src/components/molecules/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { axe } from 'jest-axe';
+import React from 'react';
+import { render } from '@testing-library/react';
+import Tooltip from './Tooltip';
+
+describe('<Tooltip />', () => {
```
lluia: I didn't think necessary to add more tests, given we're just wrapping a library that is quite well tested already 👍🏻

PR #165, opened 2020-01-15
Fix <Tooltip /> CSS 🔧

lluia:
Following #164, get [Tippy's CSS from a CDN](https://atomiks.github.io/tippyjs/getting-started/#2-cdn) rather than add it as part of our package.

This is to keep our build simple [and passing](https://app.circleci.com/jobs/github/zopaUK/react-components/1312) without the need to make **Rollup** aware of how to handle CSS files 💆🏻‍♂️


package.json
```diff
@@ -89,6 +89,7 @@
     "rollup-plugin-url": "^3.0.0",
     "semantic-release": "^15.13.12",
     "styled-components": "^4.4.0",
+    "tippy.js": "^5.1.4",
```
lluia: This is already in included in **Tippy** React wrapper ( `@tippy.js/react` ) which is already declared as part of our `"dependencies"`.

Why add it to `"devDependencies"` then? Well, in order to [import the correct version from the CDN](https://github.com/zopaUK/react-components/blob/924e401df85d26d58e3820c10faf35dab2bd39ad/src/components/styles/GlobalStyles.tsx#L33-L34) we need to somehow be able to read the current version of **Tippy JS** used, hence this "hack".

src/components/styles/GlobalStyles.tsx
```diff
@@ -28,11 +29,17 @@ const imagery = css`
   }
 `;
 
+const tooltipGlobalStyles = css`
+  @import url('https://unpkg.com/tippy.js@${tippyVersion}/dist/tippy.css');
+  @import url('https://unpkg.com/tippy.js@${tippyVersion}/animations/shift-toward.css');
+`;
+
```
lluia: Adding it in `<GlobalStyles />` like this is included and run once in the client `<App />` component.

If I add it as part of `<Tooltip />` it might get run multiple times ( due to multiple instances of `<Tooltip />` and re-renders ) and cause strange side effects... 💆🏻‍♂️

package.json
lluia: Ahhh yez good point, let me try 👍

lluia: I'm actually reading from node_modules, so thinking this back there's no need to add it to `devDependencies`: it will always be there because `@tippy.js/react` installs it.

lluia:
Seems that we need to teach **Rollup** how to read JSON as well:
```
/Users/lluis.agusti/Code/zopa-react-components/src/index.ts → cjs, es...
[!] (plugin babel) SyntaxError: /Users/lluis.agusti/Code/zopa-react-components/node_modules/tippy.js/package.json: Unexpected token, expected
 ";" (2:8)

  1 | {
> 2 |   "name": "tippy.js",
    |         ^
  3 |   "version": "5.1.4",
  4 |   "description": "Highly customizable tooltip and popover library",
  5 |   "main": "dist/tippy.cjs.js",
node_modules/tippy.js/package.json (2:8)
```
( when running `yarn compile` 😪  , breaks due to [this `import`](https://github.com/zopaUK/react-components/blob/3db4ba8f570d9a617c2f6f47779a642b1178c239/src/components/styles/GlobalStyles.tsx#L4) )

PR #167, opened 2020-01-15
Re-skin <Badge /> and <AlertBox /> ( WIP )

lluia:
## Summary

New styling... 💅🏻 , as you can see now `<AlertBox />` and one version of `<Badge />` use the same icon. I made an abstraction, `<RoundIcon variant />` to handle the case 🤞🏻

### DIFF
--- a/src/components/icons/RoundIcon/RoundIcon.md
+++ b/src/components/icons/RoundIcon/RoundIcon.md
@@ -1,0 +1,18 @@
+### Summary
+
+...
+
+### Example
+
+```js
+import { RoundIcon, FlexRow, FlexContainer } from '@zopauk/react-components';
+
+<FlexContainer>
+  <FlexRow>
+    <RoundIcon />
+    <RoundIcon variant="confirmed" />
+    <RoundIcon variant="invalid" />
+    <RoundIcon variant="waiting" />
+  </FlexRow>
+</FlexContainer>;
+```
--- a/src/components/icons/RoundIcon/RoundIcon.tsx
+++ b/src/components/icons/RoundIcon/RoundIcon.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import styled from 'styled-components';
+import { colors, typography, Text } from '../../..';
+import CheckMark from '../CheckMark/CheckMark';
+
+interface IRoundIconProps {
+  variant?: 'confirmed' | 'waiting' | 'invalid';
+}
+
+interface IRoundIconCircleProps {
+  bgColor:
+    | typeof colors.neutral.nearDark
+    | typeof colors.semantic.success
+    | typeof colors.semantic.alert
+    | typeof colors.semantic.error;
+}
+
+const RoundIconCircle = styled.div<IRoundIconCircleProps>`
+  display: flex;
+  justify-content: center;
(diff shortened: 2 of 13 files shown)


src/components/icons/RoundIcon/RoundIcon.md
```diff
@@ -0,0 +1,18 @@
+### Summary
+
+...
```
lluia: **TODO:** write the docs 😆

src/components/icons/RoundIcon/RoundIcon.tsx
```diff
@@ -0,0 +1,60 @@
...
+  align-items: center;
+  background-color: ${({ bgColor }) => bgColor};
+  font-weight: ${typography.weights.bold};
+  border-radius: 100%;
+  width: 24px;
+  height: 24px;
+`;
+
+export default function RoundIcon({ variant }: IRoundIconProps) {
+  if (variant === 'confirmed')
```
poteirard: if if if -> switch?

src/components/molecules/RadioField/RadioField.md
```diff
@@ -42,7 +42,7 @@ import { RadioField } from '@zopauk/react-components';
 ```jsx
 import { RadioField } from '@zopauk/react-components';
 
-<RadioField label="I'm checked by default" inputProps={{ value: 'radio4', name: 'radio4', defaultChecked: true }} />;
```
poteirard: I'm seeing in the console:
```Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.```

File: src/components/molecules/Tooltip/Tooltip.md [hunk 4]
```diff
@@ -9,15 +9,15 @@ For more information:
...
 
-⚠️ &nbsp; We don't allow to customise `theme`, `animation` and `flipOnUpdate` to make sure the tooltip UX stays the same.
+⚠️ &nbsp; We don't allow to customise `theme`, `animation` and `flipOnUpdate` to make sure the tooltip UX stays consistent.
 
 ⚠️ &nbsp; By default you won't be able to pass JSX to the tooltip content. You can change that via `allowHTML={true}` prop.
 
 ### Example
 
 ```jsx
 import { forwardRef } from 'react';
```
poteirard: this import is not being used

src/components/icons/RoundIcon/RoundIcon.md
poteirard: Shall we wait then?

src/components/molecules/RadioField/RadioField.md
lluia: 🤔 I didn't think `defaultChecked` was a thing, I'll bring it back 👍🏻

lluia:
@poteirard I need to revert the style updates on `<Badge />`, as some teams are depending still on the old styles 😑 , I'll mark this as **WIP** again 👍🏻

lluia:
It seems that design-wise, these updates are no longer relevant 🤷🏻‍♂️

PR #169, opened 2020-01-16
Add htmlFor for label elements

moshie:
Label require a htmlFor attribute.


src/components/atoms/Text/Text.tsx
```diff
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 import { typography } from '../../../constants/typography';
 import { colors, INeutralColorSpec, ISemanticColorSpec } from '../../../constants/colors';
 
-export interface ITextProps extends HTMLAttributes<HTMLSpanElement> {
```
poteirard: [Question] is there a reason why we don't use directly a Label component? Like this one https://zopauk.github.io/react-components/#/Components/Atoms/InputLabel ?

PR #170, opened 2020-01-17
fix(inputlabel): remove text component from label

moshie:


src/components/atoms/InputLabel/InputLabel.tsx
```diff
@@ -1,21 +1,53 @@
...
+  mb?: boolean;
+  /**
+   * Accepts a subset of the Zopa brand colors. Same as the ones accepted by `<Heading />`.
+   */
+  color?:
+    | INeutralColorSpec['white']
+    | INeutralColorSpec['nearDark']
+    | INeutralColorSpec['dark']
+    | ISemanticColorSpec['success']
+    | ISemanticColorSpec['error'];
```
lluia: @moshie lets just leave `<InputLabe />` with the `htmlFor` prop ( _CSS looks good_ 👍  )

Although previously you could do something like `<InputLabe size="small" />`, it didn't have any effect as `<Text />` props were locked through the `attrs` syntax: `styled(Text).attrs({})`

lluia:
We'll need to update the snapshots as well later 👍🏻 ( `yarn test -u` )

src/components/atoms/InputLabel/InputLabel.tsx
moshie: @lluia Just to double check are we still keeping the margin bottom and color?

Like this?:

```jsx
export interface IInputLabelProps extends HTMLAttributes<HTMLLabelElement> {
  /**
   * Define the associated input identifier
   */
  htmlFor: string;
}

const InputLabel = styled.label<IInputLabelProps>`
  display: block;
  margin: 0;
  letter-spacing: 0;
  color: ${colors.neutral.dark};
  margin-bottom: 24px;
  line-height: ${typography.lineHeights.text};
  font-weight: ${typography.weights.semibold};
  font-size: 'base';
`;
```

moshie:
Test snapshots incoming waiting on Xcode to install

PR #174, opened 2020-01-21
Make form validation async

thegrinder:
Make it possible for the `validate` `<Form />` prop to be async

### DIFF
--- a/src/components/hooks/useForm/useForm.ts
+++ b/src/components/hooks/useForm/useForm.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useMemo } from 'react';
+import { useState, useCallback, useEffect, useMemo } from 'react';
 
 type TValue = any;
 export type TValues = Record<string, TValue>;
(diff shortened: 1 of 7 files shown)


src/components/hooks/useForm/useForm.ts
```diff
@@ -26,16 +26,20 @@ export interface IUseFormValues {
...
     if (validate) {
-      const errors = validate(formValues);
+      const errors = await validate(formValues);
       updateErrors(errors);
     }
   };
 
+  useEffect(() => {
+    runValidation(values);
+  }, []);
```
lluia: Should we watch for `values` here? ( _to make sure validation is re-run when they update_ ) 💅🏻

thegrinder: The validation runs `onChange` of the input so no need to. This is here just to run the validation on mount in case you the info is prefilled. Since the validation can by async now, I can't initialize the errors state with the result of `validate` anymore

PR #175, opened 2020-01-21
fix(footer): footer links base url update

### LINKED ISSUE #173: Hardcoded urls in the footer
The footer has links which are hardcoded to www.zopa.com.

In shopfront QA uses the footer in UAT to navigate through the site however, they get redirected to the production site.

moshie:
fix #173


src/components/molecules/ZopaFooter/Links/Links.tsx
```diff
@@ -136,24 +140,24 @@ const linkGroups = [
...
-        href: 'https://www.zopa.com/sitemap',
+        href: `${baseUrl}/sitemap`,
         label: 'Sitemap',
       },
     ],
   },
 ];
 
-const Links = () => (
+const Links: FC<ILinksProps> = ({ baseUrl = 'https://www.zopa.com' }) => (
```
thegrinder: no need to use `FC` here, since it would mean that Links can accept `children` which they don't.

src/components/molecules/ZopaFooter/ZopaFooter.tsx
```diff
@@ -20,18 +20,22 @@ export interface IFooterProps extends HTMLAttributes<HTMLDivElement> {
...
    */
   legalOnly?: boolean;
+  /**
+   * This sets the base url for all links in the footer
+   */
+  baseUrl?: string;
 }
 
-const ZopaFooter: React.FC<IFooterProps> = ({ legalOnly, ...rest }) => (
+const ZopaFooter: React.FC<IFooterProps> = ({ legalOnly, baseUrl = 'https://www.zopa.com', ...rest }) => (
```
thegrinder: I see that we use default props below so it's probably better to do it one way or the other. I know it's been there before but I think you can remove `FC` from here too. 

`const ZopaFooter = ({ legalOnly = false, baseUrl = 'https://www.zopa.com', ...rest }) => (...)`

PR #176, opened 2020-01-24
Move and expose form hooks and reexport their types

thegrinder:
- Move `FormContext`, `useFormContext`, and `useFieldContext` to `hooks` folder
- Export these hooks as they might be needed to create custom input components (e.g. to add `onChange`, you need to grab and call the original one you get from context through `useFieldContext`)
- Export hook types (all of them had to be reexported because of this https://github.com/facebook/create-react-app/issues/6054)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -58,7 +58,7 @@ export { default as ProfileIcon } from './components/icons/Profile/Profile';
 
 // Hooks
 export { default as useAccordion } from './components/hooks/useAccordion/useAccordion';
-export { default as useForm } from './components/hooks/useForm/useForm';
+export * from './components/hooks/useForm';
 
 export { colors, typography };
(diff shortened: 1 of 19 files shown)


src/index.ts
```diff
@@ -58,7 +58,7 @@ export { default as ProfileIcon } from './components/icons/Profile/Profile';
 
 // Hooks
 export { default as useAccordion } from './components/hooks/useAccordion/useAccordion';
-export { default as useForm } from './components/hooks/useForm/useForm';
+export * from './components/hooks/useForm';
```
lluia: We need to check with @poteirard whether we're gonna hit any issues doing this 🧐

I recall when you tried @thegrinder the types could be imported nah?

lluia:
Looking good to me  💯  , lets wait to see what @poteirard thinks about the `*` export on `index.ts`

PR #181, opened 2020-01-28
Expose errors and touched objects

thegrinder:
Expose `errors` and `touched` object for the following reasons:
- being able to grab `errors`  is gonna be helpful if we wanna do partial validation
- `touched`  is handy when we want to know when to show the error on a compound input (e.g. date of birth)


src/components/organisms/Form/hooks/useForm.ts
```diff
@@ -3,6 +3,7 @@ import { useState, useCallback, useEffect, useMemo } from 'react';
 type TValue = any;
 export type TValues = Record<string, TValue>;
 export type TErrors = Record<string, string | undefined>;
+export type TTouched = Record<string, boolean>;
```
lluia: Following your PR [on renaming these types](#196), should we call this `TFormTouched` then?

thegrinder: Since it might be considered breaking change, I'm gonna hold off with renaming for now

thegrinder: addresses here https://github.com/zopaUK/react-components/pull/207

PR #198, opened 2020-02-02
fix(link): remove padding and border from link if used as button

thegrinder:
Remove default button styling when `<Link as="button" />`


lluia:
@thegrinder should we have this as part of [`3.0`](https://github.com/zopaUK/react-components/milestone/2) ?

thegrinder:
@lluia yup

PR #204, opened 2020-02-05
Withstep progresscolor progress update

Blimeys:
Added two new optional props to the `<Progress/>` component.
**progressColor**: type of `colors`, optional, will default to previous value, allows to change the color of the progress bar - not a breaking change
**withStep**: boolean, optional, allows to toggle the step text - not a breaking change

Screenshot 2020-02-05 at 14 07 30


src/components/molecules/Progress/Progress.test.tsx
```diff
@@ -10,4 +11,13 @@ describe('<Progress />', () => {
     const results = await axe(container.innerHTML);
     expect(results).toHaveNoViolations();
   });
+  it('renders the component with the step text', () => {
+    const { queryAllByText } = render(<Progress totalSteps={5} currentStep={2} />);
+    expect(queryAllByText('Step 2 of 5')).toBeTruthy();
+    expect(queryAllByText('Step 2 of 5')).toHaveLength(2);
```
poteirard: why length 2 and without it length 1? 🤔

src/components/molecules/Progress/Progress.tsx
```diff
@@ -28,7 +30,7 @@ const SProgression = styled.div<IProgressionStyleProps>`
   border-radius: 100px;
   height: 4px;
   display: block;
-  background: ${colors.semantic.alert};
+  background: ${({ progressColor }) => (progressColor ? progressColor : colors.semantic.alert)};
```
poteirard: ```background: ${({ progressColor = colors.semantic.alert}) => progressColor};```

src/components/molecules/Progress/Progress.test.tsx
Blimeys: I think this might be a jest/snapshot issue. In that case it detects a duplicate (some form of <span>).

Blimeys: Found the culprit : `const results = await axe(container.innerHTML);` is not unmounting from the jest dom at the end of that step.

src/components/molecules/Progress/Progress.tsx
```diff
@@ -41,12 +43,14 @@ const SProgression = styled.div<IProgressionStyleProps>`
...
 
-const Progress: React.FC<IProgressProps> = ({ totalSteps, currentStep, ...rest }) => (
+const Progress: React.FC<IProgressProps> = ({ totalSteps, currentStep, withStep = true, progressColor, ...rest }) => (
   <SProgressBar {...rest}>
-    <SProgression width={`${(100 / totalSteps) * currentStep}%`}>
-      <Text size="small">
-        Step {currentStep} of {totalSteps}
-      </Text>
+    <SProgression width={`${(100 / totalSteps) * currentStep}%`} progressColor={progressColor}>
+      {withStep && (
```
thegrinder: Is `withStep` really necessary? You can easily do it outside of the component. If we add this, then the next thing you know you'll have `withPercentage`, `withSlashStepFormat`, etc.

Blimeys: The text below the bar is optional, new designs popping up dont have the text below it, i'd agree we could add up to it. I can in a separate PR if you are keen (and also do conditional types validation on props provided) but its quite a bit more work. Let me know.

src/components/molecules/Progress/Progress.test.tsx
Blimeys: I will write an issue separately to address this.

PR #209, opened 2020-02-10
Forward refs in inputs

thegrinder:
Forward refs to all `<input />` and `<select />` elements to be able to, for example, programatically focus on an element.

### DIFF
--- a/src/components/atoms/Dropdown/Dropdown.tsx
+++ b/src/components/atoms/Dropdown/Dropdown.tsx
@@ -19,7 +19,7 @@ export interface IDropdownProps extends React.SelectHTMLAttributes<HTMLSelectEle
 
 export const Option = styled.option``;
 
-const Dropdown = styled.select<IDropdownProps>`
+const StyledDropdown = styled.select<IDropdownProps>`
   appearance: none;
   background: transparent url(${chevronDown}) no-repeat calc(100% - 13px) center;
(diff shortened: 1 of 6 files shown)


src/components/atoms/Dropdown/Dropdown.tsx
```diff
@@ -38,6 +38,8 @@ const Dropdown = styled.select<IDropdownProps>`
 `;
 
 // TODO: Styleguidist to be able to locate styled components. See #147.
-export const StyleguidistDropdown: FC<IDropdownProps> = props => <Dropdown {...props} />;
+export const StyleguidistDropdown = forwardRef<HTMLSelectElement, IDropdownProps>((props, ref) => (
```
lluia: We'll need to wrap `<Dropdown />` ( _default export below_ ) instead of this one as is the one exposed by the library 👍🏻

thegrinder: done

lluia:
@thegrinder something weird with your first **5 commits**, they seem to duplicate?

Anyway you'll need to rebase and add `fix` instead of `improvement` so that[ semantic-release](https://github.com/semantic-release/semantic-release) can actually pick it and do a patch release 👍🏻

thegrinder:
@lluia not duplicates, the same change for different components

PR #211, opened 2020-02-10
Split form component

thegrinder:
Split `<Form />` into `<Form />` component that holds the state and serves as a context provider, and `<Form.Form />` that grabs `handleSubmit` off of the context and passes it to the `<form />`'s `onSubmit`

### DIFF
--- a/src/components/organisms/Form/Form/Form.md
+++ b/src/components/organisms/Form/Form/Form.md
@@ -1,6 +1,6 @@
 ### Summary
 
-`<Form>` is just a small wrapper component around the `useForm` hook. It holds the form state and passes it down through context to other [Form](#/Organisms/Form) components.
+`<Form.Form>` is just a small wrapper component around the native `form` element. It grabs `handleSubmit` off of the context and passes it to the `onSubmit` prop.
 
 ### Example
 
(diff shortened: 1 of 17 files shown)


src/components/organisms/Form/Form/Form.md
```diff
@@ -1,16 +1,18 @@
...
 
 ```js
 import { Form } from '@zopauk/react-components';
 
 <Form initialValues={{ firstName: '' }} onSubmit={values => alert(JSON.stringify(values))}>
-  <div style={{ marginBottom: '16px' }}>
-    <Form.TextField label="First name" name="firstName" />
-  </div>
-  <Form.Button>Submit</Form.Button>
+  <Form.Form>
```
lluia: Can we name this `<Form.Wrap />` or something on the lines? Seems a bit confusing `<Form.Form />` to me 💅🏻

lluia:
@thegrinder can we actually [document the breaking change](https://semantic-release.gitbook.io/semantic-release/#commit-message-format) on the commit message? If you can document [this one as well](https://github.com/zopaUK/react-components/pull/213) would be awesome ( _my bad I merged without suggesting it there_ )
```
fix(form): split form into form provider and form element

BREAKING CHANGE: <Form /> is now a top-level component, you now use it like:

BREAKING CHANGE: <Form /> validation is no more async, hence no need to...
```
In this way our [CHANGELOG](https://github.com/zopaUK/react-components/blob/master/CHANGELOG.md) is gonna be updated properly when we do the release 💅🏻

src/components/organisms/Form/Form/Form.md
thegrinder: Actually this is the biggest doubt I have here. In the end it is wrapping a native form element... not sure if `<Form.Wrap />` explains what the the component does well enough

PR #231, opened 2020-02-17
Remove input props

thegrinder:
Remove `inputProps` from all input components and spread the props directly over the inputs

### DIFF
--- a/src/components/molecules/CheckboxField/CheckboxField.tsx
+++ b/src/components/molecules/CheckboxField/CheckboxField.tsx
@@ -82,7 +82,7 @@ const Label = styled(InputLabel)`
   }
 `;
 
-const CheckboxField: React.FC<IField> = props => {
+const CheckboxField: React.FC<IField & IInput> = props => {
   const { label, name, errorMessage, inputSize, className, ...rest } = props;
 
   if (!name) throw Error('Name must be set. Check the docs.');
--- a/src/components/molecules/CheckboxField/CheckboxField.tsx
+++ b/src/components/molecules/CheckboxField/CheckboxField.tsx
@@ -83,14 +83,13 @@ const Label = styled(InputLabel)`
 `;
 
 const CheckboxField: React.FC<IField> = props => {
-  const { label, inputProps, errorMessage, size, ...rest } = props;
-  const { name } = inputProps;
+  const { label, name, errorMessage, inputSize, className, ...rest } = props;
 
-  if (!name) throw Error('Name must be set in inputProps. Check the docs.');
+  if (!name) throw Error('Name must be set. Check the docs.');
 
   return (
-    <SizedContainer size={size} {...rest}>
-      <Input id={`checkbox-id-${name}`} type="checkbox" {...inputProps} />
+    <SizedContainer size={inputSize} className={className}>
+      <Input id={`checkbox-id-${name}`} type="checkbox" {...rest} />
       <Label htmlFor={`checkbox-id-${name}`}>{label}</Label>
       {errorMessage && <ErrorMessage>{errorMessage}</ErrorMessage>}
     </SizedContainer>
--- a/src/components/molecules/CheckboxField/CheckboxField.tsx
+++ b/src/components/molecules/CheckboxField/CheckboxField.tsx
@@ -82,7 +82,9 @@ const Label = styled(InputLabel)`
   }
 `;
 
-const CheckboxField: React.FC<IField & IInput> = props => {
+export interface ICheckboxFieldProps extends IField, IInput {}
+
+const CheckboxField: React.FC<ICheckboxFieldProps> = props => {
   const { label, name, errorMessage, inputSize, className, ...rest } = props;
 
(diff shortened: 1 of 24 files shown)


src/components/molecules/DropdownField/DropdownField.test.tsx
```diff
@@ -9,7 +9,7 @@ describe('<DropdownField />', () => {
       <DropdownField
         label="Your cool dropdown ❤"
         errorMessage="You need to choose something!"
-        size="short"
+        containerSize="short"
```
endymion1818: 👍 great, it really bugged me that `size` was output to the dom

package.json
```diff
@@ -70,7 +70,7 @@
     "eslint-plugin-prettier": "3.1.2",
     "eslint-plugin-react": "7.18.3",
     "gh-pages": "2.2.0",
-    "husky": "3.1.0",
+    "husky": "4.2.3",
```
Blimeys: Can you keep bumping versions out of components work ? Easier to revert in case of 💥

src/components/molecules/DropdownField/DropdownField.tsx
```diff
@@ -29,11 +29,11 @@ const Help = styled(Text)`
 `;
 
 const DropdownField = forwardRef<HTMLSelectElement, IDropdownFieldProps>(
-  ({ label, errorMessage, size, helpText, htmlSelectSize, name, ...rest }, ref) => (
+  ({ label, errorMessage, containerSize, helpText, htmlSelectSize, name, ...rest }, ref) => (
     <>
       {label && <Label htmlFor={`text-id-${name}`}>{label}</Label>}
       {helpText && <Help size="small">{helpText}</Help>}
-      <SizedContainer size={size}>
+      <SizedContainer size={containerSize}>
```
Blimeys: Container seems to be the Fragment line 33, should wrap both label and help, this is the dropdownContainer

package.json
lluia: I think this PR is no longer in sync with `v3` as I synced that branch with `master` now.
We'll need to clean up the history 👍🏼

src/components/molecules/DropdownField/DropdownField.tsx
lluia: The purpose of this prop is to size the dropdown but not the accompanying **label** and **help text** ( _the behaviour we had before and I think this should stay as it's_ ).

This change is about renaming the prop so it doesn't collide with the native `size` HTML attribute of `<select />`.
We can rename it to `dropdownSize` if we think that's less confusing?

lluia:
@Blimeys I upgraded **Prettier** [some time ago](https://github.com/zopaUK/react-components/pull/192), I can open a PR running `yarn lint --fix` in all the files ( _be default it only does it on the diff, hence the updates on this PR_ )

src/components/molecules/DropdownField/DropdownField.tsx
Blimeys: `dropdownSize` would make more sense, I'd expect `containerSize` to be the prop for the actual wrapping container and not a prop passed down to a nested component

thegrinder: It's `containerSize` everywhere else so I wanted to make it consistent

src/components/molecules/CheckboxField/CheckboxField.tsx
```diff
@@ -82,15 +82,16 @@ const Label = styled(InputLabel)`
...
+const CheckboxField: React.FC<ICheckboxFieldProps> = props => {
+  const { label, name, errorMessage, inputSize, className, ...rest } = props;
+
+  if (!name) throw Error('Name must be set. Check the docs.');
 
   return (
-    <SizedContainer size={size} {...rest}>
-      <Input id={`checkbox-id-${name}`} type="checkbox" {...inputProps} />
+    <SizedContainer size={inputSize} className={className}>
+      <Input id={`checkbox-id-${name}`} type="checkbox" {...rest} />
```
lluia: We're then are not adding `name` anymore to this `<Input />`( _[see snapshot diff](https://github.com/zopaUK/react-components/pull/231/files#diff-7682ae740e2ddaa1075bfe3d7b94806f)_ ) is it expected?

src/components/molecules/TextField/TextField.tsx
```diff
@@ -66,15 +64,15 @@ const TextField = forwardRef<HTMLInputElement, ITextFieldProps>(
         isValid={isValid}
         aria-label={label ? undefined : name}
         ref={ref}
```
lluia: We're again not passing `name` to the `<input />`, double checking it's expected 👍🏻

src/components/molecules/CheckboxField/CheckboxField.tsx
thegrinder: fixed

src/components/molecules/TextField/TextField.tsx
thegrinder: fixed

Issue #208, opened 2020-02-07
Headings as span elements


moshie:
We have a use case whereby we need to use the styling of a heading however, semantically it does not make sense as a `h2`.

We propose that we add span to the as type under `IStyledHeadingProps`. 

https://github.com/zopaUK/react-components/blob/9ef3191c343e28b61ae0393b096c6bf22fb4f9f6/src/components/atoms/Heading/Heading.tsx#L15

![Screenshot 2020-02-07 at 10 31 09](https://user-images.githubusercontent.com/3974301/74022237-028d7080-4995-11ea-8dad-0eed58f2f49b.png)

lluia:
To me this could work 👍🏻  

One of the challenges we'll hit though is that currently, the **tag defines the size**:

https://github.com/zopaUK/react-components/blob/master/src/components/atoms/Heading/Heading.tsx#L29

So when telling it to render a `span`, you'll need an extra prop to specify which size you want:
```jsx
<Heading as="span" size="h2">Foo</h2>
```

moshie:
Might be a good idea to separate concerns?
`<Heading as="span" size="h2">Span Heading styled as a h2</Heading>`
But still have the ability to do:
`<Heading as="h2">Heading styled as a h2</Heading>`
So it doesn't break the api?

WDYT?

thegrinder:
I think we could just map the font size to `size` prop instead of `as` and let the `as` prop behave the same way as it does in any `styled` component. That would be a breaking change but there is already `v3` branch that you can create a PR against. Also, one thing to keep in mind is that if you use it as span, it will be `display: inline` by default which is different than headings which is `display: block`. So we would have to throw in an extra rule so that the behavior is consistent regardless of what element you use

moshie:
> That would be a breaking change

How come? Couldn't we implement some logic to set the size based on the as prop in the event the size prop was not passed?

Happy to implement a PR to demo?

> if you use it as span, it will be display: inline by default

Good point, I think consistency is key and the span should be `display: block;` similar to h1-6

thegrinder:
When I said it would be a braking change, I was referring to my proposed solution which means using required `size` prop instead of `as`. This way we wouldn't have to assign a default font size when the heading is `span`. On the other hand, what you're suggesting forces us to do so but you're right, it doesn't introduce any breaking changes. Not sure which would be better...

thegrinder:
[...]

PR #237, opened 2020-02-20
<Heading /> as `span` 🍵

lluia:
## Summary

Adding the ability to render heading as a `span` element:
```jsx
<Heading as="span">Foo</Heading> // by default renders at H1 size
```
and to have granular control of the rendered size through the new `size` prop:
```jsx
<Heading as="h1">Bar</Heading> // renders at 48px
<Heading as="h1" size="h6">Baz</Heading> // renders at 20px
<Heading as="span" size="h2">Bam</Heading>  // renders at 28px as a `span`
``` 

## Story
https://github.com/zopaUK/react-components/issues/208


src/components/atoms/Heading/Heading.tsx
```diff
@@ -26,8 +32,10 @@ interface IStyledHeadingProps {
 const Heading = styled.h1<IStyledHeadingProps>`
   color: ${({ color = colors.neutral.dark }) => color};
   
-  font-size: ${({ as }) => headingSizes[as]};
+  font-size: ${({ as, size }) => (as === 'span' || size ? headingSizes[size || 'h1'] : headingSizes[as])};
```
thegrinder: defaulting to `h1` size?

lluia: Any other preference, maybe we should default to `h4` or `h6` instead? 💭

thegrinder: I think I'd rather make `size` prop required and default `as` to `size` if it's not provided. Breaking change but probably a bit more elegant

lluia: I see having a different twist 🤔 , I guess is more elegant as you say however it might tricky to update for clients ( _given `as` is also used in `<Text />` hence find and replace will end messy_ )

lluia: I default it to **h4 size**  ( _which is less aggressive visually_ ) for now 💄 👍🏻

I think given there's also the icons update I prefer to keep this as not a _breaking change_, for now, to remove upgrade hassle on the clients 🙈

Issue #245, opened 2020-02-27
Build <Tooltip /> using PopperJS 🏗


lluia:
## Summary

Currently our [`<Tooltip />`](https://github.com/zopaUK/react-components/blob/master/src/components/molecules/Tooltip/Tooltip.tsx) component is just a tiny wrapper around [`@tippy.js/react`](https://github.com/atomiks/tippy.js-react#readme) with styles and animation set.

@thegrinder pointed out fairly that we could build ourselves the styles using [Popperjs](https://popper.js.org/) ( _which is what **TippyJS** uses internally_ ) and hence remove a dependency layer.

We would need to be careful to preserve the current behaviour and styling ( _see related PRs_ )

## Related

https://github.com/zopaUK/react-components/pull/164
https://github.com/zopaUK/react-components/pull/165

Issue #234, opened 2020-02-18
Make Tippy js conditional


moshie:
In shopfront we aren't using tippy.

It pulls in external scripts which would be nice to remove.

https://github.com/zopaUK/react-components/blob/master/src/components/styles/GlobalStyles.tsx

lluia:
[Those CSS](https://github.com/zopaUK/react-components/blob/master/src/components/styles/GlobalStyles.tsx#L32-L35) files are very small, I doubt they add even a couple of bytes to our bundle, so I don't they pose any issue.

The only thing that weights is `@tippy.js/react` and that's [imported on the component file](https://github.com/zopaUK/react-components/blob/master/src/components/molecules/Tooltip/Tooltip.tsx#L2). Hence if you don't `import` thank component, thanks to 🌲  _shaking_, it should not end in your bundle.

moshie:
> Those CSS files are very small

True, just an idea every little helps ;)

endymion1818:
The content download is the smallest part of the request, we still need to make the HTTP connection and negotiate SSL. We can only have 5 concurrent requests at once with HTTP/2, the less we have the more performant we can be, which is particularly critical for optimum conversions. I'd really like us to invest time to rethink this. What options are there?

lluia:
@endymion1818 @moshie it seems as easy as to move those **CSS imports** to the component file 😉 , you're the ones more interested on this so I leave finding the best solution to you 🙃

About the page load issues, _I agree in theory_, but I'm pretty sure if you compare load times with the current version of the library [`2.20.0`](https://github.com/zopaUK/react-components/blob/master/CHANGELOG.md#2112-2020-02-18) against an [old one without the tooltips](https://github.com/zopaUK/react-components/blob/master/CHANGELOG.md#274-2020-01-13) you won't see any difference?  👀

lluia:
Addressed ☝️

PR #246, opened 2020-02-27
Move Tippy CSS to our bundle 🧵

lluia:
## Summary 

We were importing **Tippy CSS** through a [CSS import on `<GlobalStyles />`](https://github.com/zopaUK/react-components/blob/master/src/components/styles/GlobalStyles.tsx).

@moshie and @endymion1818 fairly pointed out that it increased a bit the _HTTP handshake_ on the consumers not using tooltips. 

The main issue though is that [`styled-components@5` does not support anymore CSS `@import`](https://github.com/styled-components/styled-components/issues/2995).

There's work planned on [moving to just PopperJS](https://github.com/zopaUK/react-components/issues/245).
 

## Related

https://github.com/zopaUK/react-components/pull/165
https://github.com/zopaUK/react-components/issues/234

https://github.com/styled-components/styled-components/issues/2995


lluia:
I'm thinking to base this against `master` instead given it could go out now 🤔

package.json
```diff
@@ -57,7 +56,7 @@
     "@types/jest": "^24.9.0",
     "@types/jest-axe": "^3.2.1",
     "@types/node": "12.12.27",
-    "@types/react": "^16.9.22",
+    "@types/react": "16.9.19",
```
lluia: I'm still seeing type errors related to this package, need to investigate further 🧐

etc/rollup/config.js
```diff
@@ -6,7 +6,7 @@ import url from 'rollup-plugin-url';
 import babel from 'rollup-plugin-babel';
 import commonjs from 'rollup-plugin-commonjs';
 import { terser } from 'rollup-plugin-terser';
-import json from '@rollup/plugin-json';
+import css from 'rollup-plugin-css-only';
```
poteirard: Have you tested the compile command? I think it will not work since you still have `json()` in line 34.

lluia: ugh no 🤦🏻‍♂️ , on it 👍

lluia: Hmm seems fine after removing it, however, I see this warning 🧐
```
$ rollup --config ./etc/rollup/config.js

/Users/lluis.agusti/Code/zopa-react-components/src/index.ts → cjs, es...
(!) Mixing named and default exports
https://rollupjs.org/guide/en/#output-exports
The following entry modules are using named and default exports together:
src/components/atoms/Dropdown/Dropdown.tsx
node_modules/babel-preset-react-app/node_modules/@babel/runtime/regenerator/index.js

Consumers of your bundle will have to use chunk['default'] to access their default export, which may not be what you want. Use `output.
exports: 'named'` to disable this warning
[!] (plugin css) Error: ENOENT: no such file or directory, mkdir
Error: ENOENT: no such file or directory, mkdir
```

lluia:
I'm going to change a bit the approach on this PR 💭

I think I'll follow the same API as with [`<Modal.Styles />`](https://zopauk.github.io/react-components/#/Components/Molecules/Modal) and force clients to declare explicitly `<Tooltip.Styles />` on their **top-level component**.

etc/rollup/config.js
poteirard: This happens also in master so it can be addressed in another PR

Issue #151, opened 2019-12-17
Create useViewport hook or similar


dfvalero:
FlexCol component is hiding or showing the content depending on the width the viewport, this is a cool feature, but in some cases we want conditional rendering instead of hiding or showing via CSS.

We can create a useViewport hook that exports two methods (isSmallerThan, isBiggerThan) to check the viewport width in multiple components. Also, it will be great to create a Provider to avoid adding multiple listeners to the resize event.

lluia:
Sounds a cool idea to me 💯 

At **Zopa savings** we have a lightweight implementation that looks like this:
```tsx
import throttle from 'lodash.throttle';
import { PureComponent } from 'react';
export interface IViewportState {
  height: number;
  width: number;
}

interface IViewportProps {
  children: (state: IViewportState) => JSX.Element;
}

function getViewportDimensions() {
  const width = Math.max(document.documentElement.clientWidth, window.innerWidth || 0);
  const height = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);

  return {
    height,
    width,
  };
}

export default class Viewport extends PureComponent<IViewportProps, IViewportState> {
  public state = getViewportDimensions();
  public onResize = throttle(() => window.requestAnimationFrame(this.getViewportDimensions), 300);

  public componentDidMount() {
    window.addEventListener('resize', this.onResize);
  }

  public componentWillUnmount() {
    window.removeEventListener('resize', this.onResize);
  }

  public getViewportDimensions = () => this.setState(getViewportDimensions());

  public render() {
    const { width, height } = this.state;

    return this.props.children({
      height,
      width,
    });
  }
}
```
I think however that API approach you mention, using **hooks** and **context** is more modern and performant 👍🏻 

I think generally is a better practice to _conditionally render_ rather than hiding things with CSS, in this way what's hidden is not part of the component tree and renders are faster ❤️

dfvalero:
This is what we implemented at zeosHub:

index.ts
```
export { useViewport } from './useViewport';
export { ViewPortProvider } from './provider';
```

provider.tsx
```
import React, { FC, useCallback, useEffect, useState } from 'react';
import { grid } from '@zopauk/react-components';
import { ISize, ViewportContext } from './context';

export const ViewPortProvider: FC = props => {
  const [width, setWidth] = useState(window.innerWidth);

  useEffect(() => {
    // TODO: add throttle?
    const calcInnerWidth = () => setWidth(window.innerWidth);
[...]

PR #248, opened 2020-03-05
`useViewport` hook 🎣

lluia:
## Summary

```jsx
function Example() {
  const { width } = useViewport()

  return (
    <Text>
      {width > breakpoints.tablet ? '🍅 Tomato!' : '🥔 Potato!'}
    </Text>;
  )
}
```
Exposing a new `useViewport()` utility which can be used to **programmatically read the viewport size** on a React component.

@dfvalero I kept it simple for now without `React.useContext`, let me know your thoughts 🤞🏻

## Other 🧽

- created a `src/hooks` folder
- created a **"hooks" section** on Styleguidist for documentation 💄 
- exported our breakpoints in `src/helpers/responsiveness`
- cleaned-up some dead links in the docs...

## Story

#151


src/hooks/useViewport/useViewport.ts
```diff
@@ -0,0 +1,31 @@
...
+  const height = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
+
+  return {
+    height,
+    width,
+  };
+}
+
+export function useViewport({ timeout = 300 }: IUseViewportOptions = {}) {
+  const [size, setSize] = useState(readViewport());
```
Blimeys: This might be quite heavy if someone start playing with the window size ?

dfvalero: Exact. We are generating an eventListener every time some component calls the useViewport hook. Maybe, we can improve this using a context.

lluia: @Blimeys won't be a problem as the event listener is throttled 💅🏻

@dfvalero you're right on the fact that a listener is added to `window` **every time** a mounted component uses it 🤧 . However, I was initially drawn away from forcing context on this ( _seemed a bit too much..._ 💄 ) maybe we could enable it via an option?
```jsx
/**
 * When supplying this option the hook will throw unless there's
 * a <ViewportProvider /> above on the tree. In this mode the hook
 * will "connect" to the context rather than attach its ow listener
 */
function Button() {
  const {width} = useViewport({ context: true })
  // ...
}

/**
 * Reads viewports sizes from its own listener (as it's today) 
 */ 
function Table() {
  const {width} = useViewport()
}
```

In this way, you won't have an extra context in your app unless you have a use case for it having multiple components on a single screen using this hook.

dfvalero: What about this:
- We create the Context
- if we call useViewport in a component that is a child of the Context Provider, we use the context.
- if we call useVieport in a component that is **not** a child of the Context Provider, we use a regular useEffect solution

too complicated?

lluia: @dfvalero sounds good to me 💯 , to clarify this is how it ends:
```jsx
import { useViewport, ViewportContext } from '@zopauk/react-components'

const Button = () => {
  const {width} = useViewport()
  // ...
}

const App1 = () => (
  <ViewportContext.Provider>
    <Button /> {/* <Button /> will use context given there's a provider above */}
  </ViewportContext.Provider>
) 

const App2 = () => (
  <>
    <Button /> {/* <Button /> will uses listener given ther's no context above */}
  </>
) 
```

dfvalero: beeautiiiiiful <3

src/hooks/useViewport/useViewport.md
```diff
@@ -0,0 +1,45 @@
...
+
+`useViewport()` can be used to programatically read the viewport size on a React compnent.
+
+It can be useful for conditional rendering, which is usually prefered over hiding with CSS given it reduces the VDOM tree size.
+
+### Tips
+
+By default this hook will attach an event listener to `window.resize` that will read the current viewport dimensions and let you know when those change.
+
+**In case you're using this hook in more than one component for a given page** ... please use the provided `ViewportProvider` to prevent overload `window` with `resize` listeners 🚨
```
dfvalero: <3

src/hooks/useViewport/useViewport.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    height,
+    width,
+  };
+}
+
+export function useViewport({ timeout = 300 }: IUseViewportOptions = {}) {
+  const contextSize = useContext(ViewportContext);
+  if (contextSize !== undefined) return contextSize;
```
dfvalero: eaaaasy and clean!

src/hooks/useViewport/useViewport.test.tsx
```diff
@@ -0,0 +1,47 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import { fireEvent, wait } from '@testing-library/react';
+import { useViewport } from './useViewport';
+
+describe('useViewport', () => {
```
dfvalero: what do you think about addind a test that checks that the resize method is called only once when the hook is wrapped in the Provider? too much tests?

src/helpers/responsiveness.ts
```diff
@@ -2,8 +2,7 @@ import { css } from 'styled-components';
 
 export type TDeviceSizes = 'tablet' | 'phone' | 'desktop';
 
-// TODO: Consider the breakpoints defined by google's material design https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-breakpoints
-const sizes = {
+export const breakpoints = {
```
poteirard: I think a better typing would be
```suggestion
export const breakpoints = {
  desktop: 1280,
  phone: 600,
  tablet: 720,
} as const;

export type TDeviceSizes = keyof typeof breakpoints;

type TMedia = Record<TDeviceSizes, ReturnType<typeof interpolate>>;
  
const maxMedia: TMedia = {
  phone: interpolate('phone', 'max'),
  tablet: interpolate('tablet', 'max'),
  desktop: interpolate('desktop', 'max'),
};

const minMedia: TMedia = {
  phone: interpolate('phone', 'min'),
  tablet: interpolate('tablet', 'min'),
  desktop: interpolate('desktop', 'min'),
};
```

src/hooks/useViewport/ViewportProvider.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useViewport, IUseViewportOptions } from './useViewport';
+
+interface IViewportSize {
+  width: number;
+  height: number;
+}
+
+type TViewportContext = IViewportSize | undefined;
+
+const ViewportContext = createContext<TViewportContext>(undefined);
```
poteirard: I think it will be better to just pass `undefined` as default to `width` and `height` the 
```suggestion
const ViewportContext = createContext({
  width: undefined,
  height: undefined, 
});
```
then we can rid of the `interface IViewportSize` of `type TViewportContext`

src/hooks/useViewport/useViewport.md
```diff
@@ -0,0 +1,45 @@
+### Summary
+
+`useViewport()` can be used to programatically read the viewport size on a React compnent.
+
+It can be useful for conditional rendering, which is usually prefered over hiding with CSS given it reduces the VDOM tree size.
```
poteirard: typo
```suggestion
It can be useful for conditional rendering, which is usually preferred over hiding with CSS given it reduces the VDOM tree size.
```

src/hooks/useViewport/useViewport.md
```diff
@@ -0,0 +1,45 @@
+### Summary
+
+`useViewport()` can be used to programatically read the viewport size on a React compnent.
```
poteirard:
```suggestion
`useViewport()` can be used to programmatically read the viewport size on a React component.
```

src/hooks/useViewport/useViewport.test.tsx
lluia: I'll give it a go 👍🏻

src/hooks/useViewport/useViewport.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const [size, setSize] = useState(contextSize || readViewport());
```
poteirard: If you are doing ` if (contextSize !== undefined) return contextSize;`

when this condition is going to happen? `useState(contextSize || readViewport())`

lluia: ahhh good, I think it should be here just: `useState(readViewport)` 👌🏻

poteirard:
```suggestion
  const [size, setSize] = useState(readViewport());
```

lluia:
@poteirard I guess if we do:
```ts
interface IViewportSize {
  width?: number;
  height?: number;
}
```
Then the viewport size can be `{}` or `{width: 1000, height: 5000}`. 

Seems a bit cleaner to me to keep the shape of the object and just default the values as `undefined` 🤔 💅🏻 , happy to change though if you want 👍🏻
```

File: src/hooks/useViewport/useViewport.test.tsx [hunk 10]
```diff
@@ -0,0 +1,96 @@
...
+          <Component />
+        </>,
+      );
+
+      const resizeListeners = filterListeners(window.addEventListener.mock.calls);
+
+      expect(resizeListeners.length).toEqual(2);
+    });
+
+    it('registers 1 listener for multiple instances with context', () => {
```
dfvalero: good one!

dfvalero:
all the code looks good to me, but you have some conflict files

Issue #158, opened 2020-01-08
Upgrade  `react-styleguidist@11`


lluia:
Supports TS out of the box, so we might not need anymore `react-docgen-typescript`.

See  👀
https://github.com/styleguidist/react-styleguidist/pull/1500
https://github.com/styleguidist/react-styleguidist/pull/1514/files

poteirard:
It will be present in the upcoming major version [react-styleguidist@11](https://github.com/styleguidist/react-styleguidist/milestone/18)

PR #250, opened 2020-03-07
Upgrade react styleguidist

poteirard:
- Upgrade react-stylegudist to version 11. Fixes: #158 

Note: The new version does support TS but we still need to use `react-docgen-typescript` to filter the `@types/react` and to use the `componentNameResolver`. More info can be found [here](https://github.com/styleguidist/react-styleguidist/pull/1514#discussion_r375616656)

Additional stuff:
- Use TS extension for jest `setupTests` file
- Use `ts` and `tsx` in markdowns (thanks to the new version)
- Simplify styleguidist.config


styleguide.config.js
```diff
@@ -197,8 +186,6 @@ module.exports = {
     url: 'https://github.com/zopaUK/react-components',
     text: 'Fork me!',
   },
-
-  ignore: ['**/*.test.{js,jsx,ts,tsx}', '**/*.d.ts', '**/src/styleguide-components/**', '**/styleguide/**'],
```
poteirard: this is the default, except for the `src/styleguide-components` but the components are always taken from `src/components` so it's not needed

styleguide.config.js
```diff
@@ -51,64 +51,53 @@ module.exports = {
...
       sectionDepth: 2,
       name: 'Components',
       content: 'src/components/README.md',
-      components: 'src/components/*/*.{js,jsx,ts,tsx}',
       sections: [
         {
           name: 'Atoms',
           content: 'src/components/atoms/README.md',
-          components: 'src/components/atoms/*/*.{js,jsx,ts,tsx}',
-          ignore: '**/*.test.{js,jsx,ts,tsx}',
```
poteirard: `tests` are ignored by default

Issue #139, opened 2019-11-07
Type `<Link color />` properly


lluia:
The `color` prop on `<Link />` is currently typed as `any`, however this is the [type it should point to](https://github.com/zopaUK/react-components/blob/master/src/constants/colors.ts#L2), which represents the colour values we expect to be supplied.

The reason to be typed as `any` is that it was conflicting with the `color` definition of `<Text />` ( _as it used to extend it_ ), now that `<Link />` just extends `styled.a` we might not have this problem anymore.

moshie:
Do we have too hard define the string values can we not just define it as a string?
Sometimes we find that links are colours not defined in this union type.

I can give an example in a PM.

lluia:
I think this was [fixed on *v3*](https://github.com/zopaUK/react-components/blob/master/src/components/atoms/Link/Link.tsx#L10) but I forgot to close the issue 😅 

In *v4* is gonna be even less of an issue given we [only gonna have two colours](https://github.com/zopaUK/react-components/blob/milestone/4.0.0/src/components/atoms/Link/Link.tsx).

PR #254, opened 2020-03-11
Type link color properly 🧼

lluia:
## Summary

```jsx
<Link color />
```
This prop was typed as `any` given previously this component was extending `<Text />` and **TS** was getting a mess with the `color` props.

Now that it just extends `styled.a` type it properly...
 
## Issue

https://github.com/zopaUK/react-components/issues/139

### DIFF
--- a/src/components/atoms/Link/Link.tsx
+++ b/src/components/atoms/Link/Link.tsx
@@ -1,22 +1,23 @@
 import React from 'react';
 import styled from 'styled-components';
-import { colors } from '../../../constants/colors';
+import { colors, TLinkHexColors } from '../../../constants/colors';
 import { typography } from '../../../constants/typography';
 
 export interface ILinkProps
   extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
     React.RefAttributes<HTMLAnchorElement> {
   target?: '_blank';
-  color?: any; // see issue #139 for more context on why the explicit `any` here.
+  color?: TLinkHexColors;
   as?: 'a' | 'button';
 }
 
 const SLink = styled.a<ILinkProps>`
+  background-color: transparent;
   font-size: inherit;
   font-family: ${typography.primary};
   font-weight: ${typography.weights.semibold};
   line-height: ${typography.lineHeights.text};
-  color: ${({ color = colors.neutral.dark }) => color};
+  color: ${({ color }) => color};
   cursor: pointer;
   text-decoration: none;
   user-select: auto;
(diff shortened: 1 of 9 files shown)


src/components/atoms/Link/Link.tsx
```diff
@@ -1,22 +1,23 @@
...
   extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
     React.RefAttributes<HTMLAnchorElement> {
   target?: '_blank';
-  color?: any; // see issue #139 for more context on why the explicit `any` here.
+  color?: TLinkHexColors;
   as?: 'a' | 'button';
 }
 
 const SLink = styled.a<ILinkProps>`
+  background-color: transparent;
```
lluia: Some browsers apply a `background-color` by default when a `button` is rendered, normalised that here 💆🏻‍♂️

src/constants/colors.ts
```diff
@@ -1,5 +1,5 @@
 export type TTextHexColors = '#FFFFFF' | '#63637E' | '#0D0A38' | '#1EC06A' | '#EE0505';
-export type TLinkHexColors = '#FFFFFF' | '#0D0A38';
+export type TLinkHexColors = '#FFFFFF' | '#4A3EDE';
```
lluia: This type wasn't used and hence was pointing to the wrong colour... 🎨

lluia:
I tested the components locally and all display the proper color on links 👍🏻 💆🏻‍♂️

src/components/atoms/Link/Link.tsx
```diff
@@ -1,22 +1,23 @@
...
-import { colors } from '../../../constants/colors';
+import { colors, TLinkHexColors } from '../../../constants/colors';
 import { typography } from '../../../constants/typography';
 
 export interface ILinkProps
   extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
     React.RefAttributes<HTMLAnchorElement> {
   target?: '_blank';
-  color?: any; // see issue #139 for more context on why the explicit `any` here.
+  color?: TLinkHexColors;
```
thegrinder: you can probably make it required now here since there is no default anymore and the fallback is provided in the wrapper

lluia: `ILinkProps` is used to type both `<Link />` and `<SLink />`, if we make it required, we would need to create a new `interface` to **type `<SLink color />` only required**: seems a bit of an overkill for the little benefit it brings?  💆🏻‍♂️

PR #269, opened 2020-03-23
fix(footer): added alt to svg social icons

moshie:


src/components/molecules/ZopaFooter/SocialLinks/SocialLinks.tsx
```diff
@@ -97,6 +101,8 @@ const socialLinksData: ISocialLinkData[] = [
           width="30px"
           viewBox="0 0 24 24"
         >
+          <title>Twitter</title>
+          <desc>bird icon for the twitter logo</desc>
```
thegrinder: Slightly late but I think we shouldn't be using word icon in the description. Kinda like with the images where you don't say `image` in the `alt` attribute

moshie: @thegrinder No worries I will make another PR :)

What do you think the Alt should be "bird for the twitter logo" ?

PR #290, opened 2020-03-30
Update Flexible container size

moshie:
I spoke to Rob and confirmed the container width and gutter size for the new designs

### DIFF
--- a/src/constants/grid.ts
+++ b/src/constants/grid.ts
@@ -13,7 +13,7 @@ const grid: IGrid = {
     l: 992,
     m: 768,
     s: 576,
-    xl: 1250,
+    xl: 1300,
     xs: 0,
   },
   cols: 12,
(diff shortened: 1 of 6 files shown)


src/constants/grid.ts
```diff
@@ -13,16 +13,16 @@ const grid: IGrid = {
     l: 992,
     m: 768,
     s: 576,
-    xl: 1200,
+    xl: 1250,
```
dfvalero: the xl sizes and the gutter size look a bit weird to me :|

moshie: Might be worth speaking to the designers about this.
This update is based on the most recent designs for shopfront.

moshie: turns out the gutter spacing is 24 not 22 so I have updated the breakpoint to 1300 which is slightly less of a random number :D

lluia:
@moshie can you change the base branch to `milestone/4.0.0` ? 

I'm thinking this has quite a potential to break layouts so I prefer to have it as part of a major release 👨🏻‍🔧 📐

PR #293, opened 2020-04-02
Typography updates 🪀

lluia:
## Summary 💭

Screenshot 2020-04-07 at 17 27 19

Screenshot 2020-04-07 at 17 27 42

Update our [typographic scale](GatsbyLink) according to the [latest designs](https://marvelapp.com/5504e6j/screen/67099627).

I had to update `<Heading />`, `<Text />` and `<Link />` however their APIs are almost the same.

Important things to note:
- every `font-size` has its matching `line-height` so generic `line-heights` anymore
- consumers will be now in charge of importing **Open Sans**

## Preview 🔍 

[Temporarily accessible](https://zopauk.github.io/react-components/beta/index.html#/Components/Atoms/Heading)

## TODO 🤿

- [x] Confirm weights loaded ( currently 4 proposed: `400|600|700|800` )
- [x] Confirm whether we use `rem` or not for `font-size`
- [ ] ~~`<Box />` component + `color: inherit` on `<Text />`~~ ( moved to next PR )
- [x] Write description ✍️

## Design
[Marvel](https://marvelapp.com/5504e6j/screen/67099627)

## Stories
[ZDS-3](https://jira.zopa/browse/ZDS-3)
[ZDS-7](https://jira.zopa/browse/ZDS-7)
[ZDS-10](https://jira.zopa/browse/ZDS-10)


src/components/atoms/Link/Link.tsx
```diff
@@ -1,36 +1,37 @@
...
-import { colors, TColors } from '../../../constants/colors';
+import { colors } from '../../../constants/colors';
 import { typography } from '../../../constants/typography';
 
 export interface ILinkProps
   extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
     React.RefAttributes<HTMLAnchorElement> {
   target?: '_blank';
-  color?: TColors['actionPlain'] | TColors['white'];
+  negative?: boolean;
```
lluia, line 10: The reason I chose a `boolean` prop rather than a `color` or `variant` one is that the **design team** wants to move away from text on top of dark background.

This is kept for legacy puprposes but probably de deprecated and removed in *v5* 💅🏻 🧽 🙇🏻‍♂️

src/components/atoms/Heading/Heading.tsx
```diff
@@ -1,45 +1,65 @@
...
+
+const lineHeightMap = {
+  display: '76px',
+  h1: '54px',
+  h2: '46px',
+  h3: '36px',
+  h4: '32px',
+  h5: '26px',
+  h6: '20px',
+};
```
lluia, line 47: The reason I inlined the `line-height` on `<Text />`, and `<Heading />`, rather than having them in our `typography` constants is that I don't see much of a use case for them to be there.

It's still useful to keep the `font-size` though in the `typography` constants as they're read in our **form controls components**.

src/styleguide-components/StyleGuide.js
```diff
@@ -9,13 +9,9 @@ import Ribbon from 'rsg-components/Ribbon';
...
 
 import GlobalStyles from '../components/styles/GlobalStyles';
-import Fonts from '../components/styles/Fonts';
 
 const GlobalStyleGuideStyles = createGlobalStyle`
-  /* System Fonts as used by GitHub */
-  body {
-    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
-  }
+   @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,600,700,800&display=swap');
```
poteirard, line 14: aren't we going to get rid of this and instead use a `<link>`?

src/components/atoms/Text/Text.tsx
```diff
@@ -6,27 +6,48 @@ import { colors, TColors } from '../../../constants/colors';
...
   mb?: boolean;
   /**
-   * Accepts a subset of the Zopa brand colors. Same as the ones accepted by `<Heading />`.
+   * Whether to render the text in all caps or not.
+   * @default false
+   */
+  capitalize?: boolean;
+  /**
+   * Where the rendered text should be aligned to.
+   * @default 'left'
```
thegrinder, line 35: Well, technically there is no default from what I see which works just fine. If you decide to add a default value, than it should be `inherit` rather than `left`

thegrinder: same thing goes for `Heading`

src/styleguide-components/StyleGuide.js
lluia: Given this is just our *Stylgeuide* docs I thought it doesn't matter much ( _as they don't use `styled-components` in "production" mode?_ )

Apps should use `<link />` though, I wonder where would be the best place to document that 🤔

src/components/atoms/Text/Text.tsx
lluia: Yup I'll update 👍

src/constants/typography.ts
```diff
@@ -5,26 +5,25 @@ const typography = {
...
   weights: {
     regular: 400,
-    semibold: 600,
+    semiBold: 600,
     bold: 700,
-  },
-  lineHeights: {
-    text: 1.6,
-    heading: 1.1,
+    extraBold: 800,
```
alexmngn, line 26: extraBold is not a thing. The maximum weight we have is 700.

src/components/atoms/Heading/Heading.tsx
alexmngn: Is there any reason why the letter spacing is not recorded?

src/components/atoms/Text/Text.tsx
```diff
@@ -6,27 +6,48 @@ import { colors, TColors } from '../../../constants/colors';
 
 export interface ITextProps extends HTMLAttributes<HTMLSpanElement> {
   /**
-   * The weight of the rendered text. Avoid the use of `semibold` as for now.
+   * The weight of the rendered text.
+   * @default 'regular'
    */
-  weight?: keyof typeof typography.weights;
+  weight?: 'regular' | 'bold';
```
alexmngn, line 12: semiBold should be supported instead of bold.

src/components/atoms/Text/Text.tsx
```diff
@@ -39,18 +60,17 @@ const Text = styled.span<ITextProps>`
     margin-bottom:   24px
   `};
 
-  line-height: ${typography.lineHeights.text};
+  font-size: ${({ size = 'body', capitalize }) => typography.sizes.text[capitalize ? 'small' : size]};
+  line-height: ${({ size = 'body' }) => lineHeightMap[size]};
+  font-weight: ${({ weight = 'regular', capitalize }) => typography.weights[capitalize ? 'bold' : weight]};
```
alexmngn, line 65: Why capitalized text is automatically bold?

src/components/atoms/Text/Text.tsx
lluia: According to @Gianeo only these two for regular text:
> Normal + bold => body text, Semibold => underlined links
I asked about it on the private rebrand channel 📐

src/constants/typography.ts
lluia: It's definitively a thing in web and we had a whole meeting about it on Friday 😁

What are you following for the source of truth on this? 

We're following what's on [Marvel](https://marvelapp.com/5504e6j/screen/67099627). and catching up with design daily 👍🏻

src/components/atoms/Heading/Heading.tsx
lluia: Good catch 👏🏻 , I'll address 👍

PR #302, opened 2020-04-20
Atomic Spacing Classes

moshie:


thegrinder:
I'm not sure if we need the components to be honest, I think just class names would be enough. If we decide to go with class names only, I think throwing in tailwindcss would remove pretty much all the logic

src/README.md
```diff
@@ -49,3 +49,24 @@ const App = () => (
...
+### Spacing
+
+```ts static
+import { Margin } from '@zopauk/react-components';
+
+// root component
+const App = () => (
+  <>
+    <Margin />
+    <Padding />
```
lluia, line 62: I suggest making this part of [`<GlobalStyles />`](https://github.com/zopaUK/react-components/blob/master/src/components/styles/GlobalStyles.tsx) instead, in this way consumers need to do less setup 💅🏻

lluia: @thegrinder also reply for yours I guess ☝️

lluia:
Thanks for this @moshie 💯 ❤️

src/README.md
moshie: I was considering that but I wasn't sure if people would like the flexibility?

thegrinder:
I think you don't need to convert everything to strings. You can probably simplify the logic if you just pass style object to `css` helper - i.e. `css({ marginBottom:  '16px' })` renders just fine

src/README.md
lluia: I think this should come yes or yes when using the library so that people stick our desing system 🙏🏻

moshie:
> I think you don't need to convert everything to strings. You can probably simplify the logic if you just pass style object to `css` helper - i.e. `css({ marginBottom: '16px' })` renders just fine

Wouldn't I need to manipulate the string to camel case to get `marginBottom` as the object key?

thegrinder:
@moshie yeah you would but you'd be able to remove  `join()`s and some string interpolations for sure

src/README.md
moshie: Done 👍

src/README.md
```diff
@@ -49,3 +49,24 @@ const App = () => (
...
+    // rest of your top-level components
+  </>
+);
+```
```
lluia, line 66: Could we remove this code example? I think it would tempt people to do it that way rather than rely on `<GlobalStyles />` 💆🏻‍♂️

I also suggest we add a **table** here with all the available `maring`and `padding` classes as a reference for people. In case that's too much work we can add a [link to tailwind](https://tailwindcss.com/docs/margin/#app)?

moshie: that was something I was planning to do this is still very much WIP.
Will update soon!

src/constants/spacing.ts
```diff
@@ -1 +1 @@
-export const spacing = [4, 8, 16, 24, 32, 40, 56, 64, 104];
+export const spacing = [0, 4, 8, 16, 24, 32, 40, 56, 64, 104];
```
thegrinder, line 1: I'm aware that this is part of the specs but we should be more granular here, as I'm pretty sure you're gonna be short a few values especially on the lower spectrum (similar to what tailwind offers in default config https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js#L130). I'm saying this cause once we start using it, and then realize we're missing a few values in between, it's gonna be difficult to upgrade

src/components/styles/Spacing.ts
```diff
@@ -0,0 +1,75 @@
...
+        .${breakpoint as TGridBreakpoints}${CSS.escape(':')}${type.substr(0, 1)}${shortHandPosition}-${sizingIndex} {
+          ${propertyFactory(type, sizingValue, position)}
+        }
+      `,
+        )}
+      }
+    `,
+    )
+    .join(' ')}
+`;
```
thegrinder, line 75: there is quite a bit of logic here and I think it would simplify things a bit and save us a few lines actually if we did some of the stuff by hand, i.e.
```
const createSpacingStyles = (type: 'm' | 's') =>
  Object.keys(grid.breakpoints).reduce(
    (mediaQueries, breakpoint) => ({
      ...mediaQueries,
      [`@media screen and (min-width: ${grid.breakpoints[breakpoint as TGridBreakpoints]}px)`]: {
        ...sizes.reduce(
          (classNames, size) => ({
            ...classNames,
            [`.${breakpoint}:${type}a-${size}`]: {
              margin: size,
            },
            [`.${breakpoint}:${type}v-${size}`]: {
              marginTop: size,
              marginBottom: size,
            },
            [`.${breakpoint}:${type}h-${size}`]: {
              marginRight: size,
              marginLeft: size,
            },
            [`.${breakpoint}:${type}t-${size}`]: {
              marginTop: size,
            },
            [`.${breakpoint}:${type}r-${size}`]: {
              marginRight: size,
            },
            [`.${breakpoint}:${type}b-${size}`]: {
              marginBottom: size,
            },
            [`.${breakpoint}:${type}l-${size}`]: {
              marginLeft: size,
            },
          }),
          {},
        ),
      },
    }),
    {},
  );
```

thegrinder: haven't tried if it works but just to give an idea

thegrinder: also it already includes the variation with spacing on all sides :)

src/constants/spacing.ts
lluia: Good point 👍🏻  probably using an **Object** is gonna give us more flexibility in the future 💆🏻‍♂️

src/components/styles/Spacing.ts
moshie: Thanks for the comments @thegrinder I have updated the code.

src/constants/spacing.ts
moshie: Updated to use an object hope

src/components/styles/Spacing.ts
```diff
@@ -0,0 +1,82 @@
+import { css } from 'styled-components';
+import { spacing as sizes } from '../../constants/spacing';
+import grid, { TGridBreakpoints } from '../../constants/grid';
+
+type TSpacingTypes = 'margin' | 'padding';
+
+const createTopLevelSizes = (type: TSpacingTypes) =>
+  Object.keys(sizes).reduce(
```
thegrinder, line 8: there is `0` breakpoint so you shouldn't need this part actually

src/components/styles/Spacing.ts
moshie: I also swapped out a, h & v as it's closer to the tailwind documentation.

src/components/styles/Spacing.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+    }),
+    {},
+  );
+
+const createResponsiveSizes = (type: TSpacingTypes) =>
+  Object.keys(grid.breakpoints).reduce(
+    (mediaQueries, breakpoint) => ({
+      ...mediaQueries,
+      [`@media screen and (min-width: ${grid.breakpoints[breakpoint as TGridBreakpoints]}px)`]: {
```
thegrinder, line 42: or you could filter it out here. up to you

src/components/styles/Spacing.ts
moshie: My bad I didn't realise there was a zero breakpoint thanks

src/README.md
```diff
@@ -49,3 +49,49 @@ const App = () => (
...
+  '0': '0',
+  '1': '4px',
+  '2': '8px',
+  '3': '16px',
+  '4': '24px',
+  '5': '32px',
+  '6': '40px',
+  '7': '56px',
+  '8': '64px',
+  '9': '104px',
```
thegrinder, line 76: with regards to my previous comment, having it as an object is one thing, I was referring more to the fact that if you decide to add a value somewhere in between, you'll have to bump the rest which. Any chance we can add a few more values at the beginning just to be sure that we don't have to go back and adjust it later down the road?

```
  '0': '0',
  '1': '4px',
  '2': '8px',
  '3': '12px',
  '4': '16px',
  '5': '20px',
  '6': '24px',
```

src/components/styles/Spacing.ts
moshie: Looking at it closer I prefer p-2 over `xs:p-2` I am going to filter zero out of the breakpoints

src/README.md
thegrinder: these will be the ones used the most. It's much easier to add or throw in something at the other end of the range as they will be used less often

PR #309, opened 2020-04-23
Add new Accordion designs

poteirard:
### 🤔 Questions for the reviewer

1. Do you guys are happy with the icons in CSS?
2. Do I have to remove the Arrow icon as it's not longer used?

### 🖌 Designs

Designs are not defined as 100% as I don't have a reference for the open state. Ask the design team is they are happy. I used this reference: https://marvelapp.com/e14bhc4/screen/67764982/handoff

SHOPFRONT_landing_page_refresh__Q2_2020_

Mine looks like this:

React_components

### DIFF
--- a/src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
+++ b/src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
@@ -1,9 +1,9 @@
 import React, { FC, HTMLAttributes } from 'react';
 import styled from 'styled-components';
 import { colors } from '../../../../constants/colors';
-import Arrow from '../../../icons/Arrow/Arrow';
 import Text from '../../../atoms/Text/Text';
 import { useAccordionContext } from '../hooks';
+import { spacing } from '../../../../constants/spacing';
 
 export interface IAccordionHeader extends HTMLAttributes<HTMLButtonElement> {
   id: string;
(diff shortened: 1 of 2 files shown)


moshie:
Do you guys are happy with the icons in CSS?
I see no issue in fact I would say it's better because you don't have extra elements to load. 👍

Do I have to remove the Arrow icon as it's not longer used?
Yes please.

lluia:
@moshie I need to open a PR removing the existing icons and using **font-awesome** ones instead ( and add the `font-awesome` import to **Styleguidist** ), is the last thing I plan to do on **ZRC** before my furlough 📐

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -14,46 +14,48 @@ export interface IAccordionHeader extends HTMLAttributes<HTMLButtonElement> {
...
+const Cross = styled.span<{ active: boolean }>`
+  width: 10px;
+  height: 10px;
+  // This background creates the ➖
+  background: linear-gradient(to bottom, transparent 35%, ${colors.grey} 35%, ${colors.grey} 65%, transparent 65%)
+    ${({ active }) =>
+      !active &&
+      // This adds the vertical stick to create ✚
+      `, linear-gradient(to right, transparent 35%, ${colors.grey} 35%, ${colors.grey} 65%, transparent 65%)`};
+`;
```
lluia, line 47: Have you tested this looks good in all browser? 💭 

Ha ha ha maybe we can replace them with a **font-awesome** SVG later when we refactor the icons 👍🏻

moshie:
> @moshie I need to open a PR removing the existing icons and using **font-awesome** ones instead ( and add the `font-awesome` import to **Styleguidist** ), is the last thing I plan to do on **ZRC** before my furlough 📐

No problem, the Arrow can still be removed

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
poteirard: Firefox, Chrome and Safari

poteirard:
I will leave the deprecated button to @lluia PR as he's going to kill all the folder

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
lluia: If it really doesn't pose a layout problem lets leave it like this then 📐

PR #311, opened 2020-04-24
update version to v4.0.0--beta.5

moshie:
Simple version bump to .5


lluia:
Can the commit be just the version name: `4.0.0--beta.5` ?
( _to stay consistent with all the previous ones, no need `fix` given we didn't patch anything_ 👍🏻 )

PR #312, opened 2020-04-24
Deprecate existing icon components 🗑

lluia:
## Summary

Screenshot 2020-04-24 at 13 19 51

Deprecate existing icon components using [`util-deprecate`](https://www.npmjs.com/package/util-deprecate) and point to which **font-awesome** icons to use instead.

## Story

[ZDS-11](https://jira.zopa/browse/ZDS-11)

### DIFF
--- a/src/components/atoms/AlertBox/AlertBox.tsx
+++ b/src/components/atoms/AlertBox/AlertBox.tsx
@@ -29,6 +29,14 @@ const AlertBox: FC<IAlertBoxProps> = ({ children, icon, ...rest }) => (
 );
 
 /**
- * @deprecated `<AlertBox />` is deprecated. It will soon be removed. Use `<Alert />` instead.
+ * @deprecated `<AlertBox />` is deprecated and will be removed in: `4.0.0`. Use [`<Alert severity="warning" />`](#/Components/Atoms/Alert) instead.
  */
-export default deprecate(AlertBox, '<AlertBox /> is deprecated. It will soon be removed. Use <Alert /> instead.');
+export default deprecate(
+  AlertBox,
+  `
+  ❗️ [@zopauk/react-components]
+  
+  <AlertBox /> is deprecated and it will be removed in the next major version: 4.0.0.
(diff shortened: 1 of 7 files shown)


src/components/atoms/AlertBox/AlertBox.tsx
```diff
@@ -29,6 +29,14 @@ const AlertBox: FC<IAlertBoxProps> = ({ children, icon, ...rest }) => (
 );
 
 /**
- * @deprecated `<AlertBox />` is deprecated. It will soon be removed. Use `<Alert />` instead.
+ * @deprecated `<AlertBox />` is deprecated and will be removed in: `4.0.0`. Use [`<Alert severity="warning" />`](#/Components/Atoms/Alert) instead.
```
dfvalero, line 32: <3

src/components/icons/Alert/Alert.tsx
```diff
@@ -19,4 +20,15 @@ const AlertIcon: React.FunctionComponent<IAlertProps> = ({ fillColor = colors.ba
...
-export default AlertIcon;
+/**
+ * @deprecated `<AlertIcon />` is deprecated and will be removed in: `4.0.0`. Use [this font-awesome 'exclamation-triangle' icon](https://fontawesome.com/icons/exclamation-triangle?style=solid) instead.
+ */
+export default deprecate(
+  AlertIcon,
+  `
+  ❗️ [@zopauk/react-components]
+  
+  <AlertIcon /> is deprecated and it will be removed in the next major version: 4.0.0.
```
dfvalero, line 31: but we can't use `<i class="fas fa-exclamation-triangle" />` in the actual version. In my opinion, it bill be better to remove the `use: <i class="fas fa-exclamation-triangle" /> instead` part.

lluia: You can if you import **font-awesome** in the client app. We don't plan to bundle it in this library and we'll suggest them to use [`react-fontawesome`](https://github.com/FortAwesome/react-fontawesome) instead.

PR #318, opened 2020-04-29
Accordion tweaks

tlaak:


src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -44,8 +45,15 @@ const AccordionHeader: FC<IAccordionHeader> = ({ children, id, index, textSize =
   const { getHeaderProps, isActiveSection } = useAccordionContext();
   const { ref, onClick: contextOnClick, ...headerPropsRest } = getHeaderProps(id, index);
 
+  const willBecomeActive = !isActiveSection(index);
+
   const handleClick = (e: MouseEvent<HTMLButtonElement>) => {
-    onClick && onClick(e);
+    const curried = onClick && onClick(e);
```
thegrinder, line 51: Not sure if we need to curry it. I don't see any use case where you would need an event object.

tlaak: I couldn't simply pass the boolean value to the onClick function and keep TypeScript happy. I wasn't able to find my way around the types.

thegrinder: this should work:
```
export interface IAccordionHeader extends Omit<HTMLAttributes<HTMLButtonElement>, 'onClick'> {
  id: string;
  index: number;
  textSize?: 'base' | 'small';
  onClick?: (willBecomeActive: boolean) => void;
}
```

src/components/organisms/Accordion/Accordion/Accordion.test.tsx
```diff
@@ -3,7 +3,13 @@ import { render } from '@testing-library/react';
 import { Accordion } from '..';
 
 describe('<Accordion />', () => {
-  it('should render aria-label', () => {
+  it('should render standard attributes but not require aria-label', () => {
```
lluia, line 6: Could we update this to just:
```js
it('should render supplied HTML attributes')
```
as the test is not testing `aria-label` is not required but just that `title` is propagated to the render ouput 💅🏻

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
lluia: Yup typing it like this should be enough as long as you don't pass the event to the handler...
```ts
const AccordionHeader = ({ onClick = () => undefined, ...rest }) => {
// ...
const handleClick = (e: MouseEvent<HTMLButtonElement>) => {
  // ...
  willBecomeActive && onClick()
}
```

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -44,8 +45,15 @@ const AccordionHeader: FC<IAccordionHeader> = ({ children, id, index, textSize =
   const { getHeaderProps, isActiveSection } = useAccordionContext();
   const { ref, onClick: contextOnClick, ...headerPropsRest } = getHeaderProps(id, index);
 
+  const willBecomeActive = !isActiveSection(index);
```
thegrinder, line 48: I think `willBecomeActive` is also a bit confusing since it's essentialy `inactive`. I would just stick to `isActive`

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
tlaak: Yeah I omitted the type and that did the trick

src/components/organisms/Accordion/AccordionHeader/AccordionHeader.tsx
tlaak: There's no easy way around the confusion here.

thegrinder: I personally would leave `isActive` in here and then you can rename it while implementing it in the app

tlaak: I updated the code and readme

PR #319, opened 2020-04-29
feat(inputrange): adding Input Range without controls

dfvalero:
I will add controls in a new Molecule (in a new PR). Still thinking in a proper name, I will go with
InputRangeWithControls for now.

Notes: This is a v3 version of the InputRange. We will add correct colors and states in v4.

Related JIRA: https://jira.zopa/browse/SHOPDEV-6087

### DIFF
--- a/src/components/atoms/InputRange/InputRange.test.tsx
+++ b/src/components/atoms/InputRange/InputRange.test.tsx
@@ -1,0 +1,32 @@
+import React, { ChangeEvent, useState } from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import InputRange from './InputRange';
+import { axe } from 'jest-axe';
+
+describe('<InputRange />', () => {
+  it('renders and updates as expected', async () => {
+    const ControlledInputRange = () => {
+      const [value, setValue] = useState(75);
+      const onChangeHandler = (event: ChangeEvent<HTMLInputElement>) => setValue(Number(event.target.value));
+
+      return <InputRange value={value} onChange={onChangeHandler} />;
+    };
+
+    render(<ControlledInputRange />);
+
+    const input = (await screen.findByRole('slider')) as HTMLInputElement;
+
+    expect(input.value).toEqual('75');
+
+    fireEvent.change(input, { target: { value: '25' } });
+
+    expect(input.value).toEqual('25');
+  });
+
+  it('renders the component with no a11y violations', async () => {
+    const { container } = render(<InputRange aria-label="slider" defaultValue={51} />);
+    const results = await axe(container.innerHTML);
+    expect(container.firstChild).toMatchSnapshot();
+    expect(results).toHaveNoViolations();
+  });
+});
--- a/src/components/atoms/InputRange/InputRange.tsx
+++ b/src/components/atoms/InputRange/InputRange.tsx
@@ -1,0 +1,125 @@
+import React, { InputHTMLAttributes, useState, ChangeEvent, useRef, useEffect } from 'react';
+import styled, { css } from 'styled-components';
+import { colors } from '../../../constants/colors';
+import { calculateTrackPosition } from './helpers';
+
+const trackHeight = 8;
+const thumbDiameter = 40;
+
+const TrackStyles = css`
+  box-sizing: border-box;
+  border: none;
+  border-radius: ${trackHeight / 2}px;
+  height: ${trackHeight}px;
+  background: ${colors.neutral.light};
+`;
+
+const ThumbStyles = css`
+  box-sizing: border-box;
+  border: none;
+  border-radius: 50%;
+  width: ${thumbDiameter}px;
+  height: ${thumbDiameter}px;
+  background: #007468;
+`;
+
+const ThumbStylesFocus = css`
+  background: ${colors.base.secondary};
+`;
+
+interface IInput extends InputHTMLAttributes<HTMLInputElement> {
+  trackPosition: number;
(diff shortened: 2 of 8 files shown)


dfvalero:
> Looks great! Focus styles are not present on Safari is the only bug I could find

Hi @endymion1818  seems that it doesn't trigger the :focus on safari (you can see it here also https://www.zopa.com/loans)

src/components/atoms/InputRange/InputRange.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(input.value).toEqual('25');
+  });
+
+  it('renders the component with no a11y violations', async () => {
+    const { container } = render(<InputRange aria-label="slider" defaultValue="51" />);
+    const results = await axe(container.innerHTML);
+    expect(container.firstChild).toMatchSnapshot();
+    expect(results).toHaveNoViolations();
+  });
```
lluia, line 31: I wonder what happens if someone does:
```jsx
<InputRange defaultValue="50" min="10" max="40" />
```
and whether we should throw an error accordingly? 🤔

src/components/atoms/InputRange/InputRange.tsx
```diff
@@ -0,0 +1,123 @@
...
+  height: ${thumbDiameter}px;
+  background: #007468;
+`;
+
+const ThumbStylesFocus = css`
+  background: ${colors.base.secondary};
+`;
+
+interface IInput extends InputHTMLAttributes<HTMLInputElement> {
+  trackPosition: number;
```
lluia, line 31: For a11y should we make `aria-label` required? 💅🏻

src/components/atoms/InputRange/InputRange.test.tsx
dfvalero: it's just a "wrapper" of a regular <input type="range" /> (https://developer.mozilla.org/es/docs/Web/HTML/Elemento/input/range).

We are not doing any magic there. So if the user puts this values, We will show a range with min=10, max=40 and value=40 (in this use case).

src/components/atoms/InputRange/InputRange.tsx
dfvalero: I didn't want to put aria-label required because you have two options to solve this a11y issue:
- Add aria-label to `<InputRange />` => `<InputRange aria-label="slider" />`
- Add a proper label:
```
<div>
    <label htmlFor="slider">Slider</label>
    <InputRange id="slider" />
</div>
```

src/components/atoms/InputRange/InputRange.tsx
```diff
@@ -0,0 +1,125 @@
...
+    border: 0;
+  }
+`;
+
+interface IInputRange extends InputHTMLAttributes<HTMLInputElement> {
+  defaultValue?: number;
+  value?: number;
+  min?: number;
+  max?: number;
+  step?: number;
```
thegrinder, line 91: no need for these here as they are already included in `HTMLInputElement` interface

dfvalero: I wanted to force it to number, to avoid "string" issues (NaN)

src/components/atoms/InputRange/InputRange.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+const InputRange = ({ min = 0, max = 100, onChange, ...otherProps }: IInputRange) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [trackPosition, setTrackPosition] = useState(0.5);
```
thegrinder, line 96: shouldn't this be initialized with `defaultValue`?

src/components/atoms/InputRange/InputRange.tsx
dfvalero: HTMLInputElement accepts min, max as string, and value as []string. I wanted to have a cleaner API

src/components/atoms/InputRange/InputRange.tsx
lluia: Good point, I forgot to ask about the `0.5` 🤔

src/components/atoms/InputRange/InputRange.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const calculateTrackPositionFromInput = (input: HTMLInputElement) => {
+    return calculateTrackPosition({ min: Number(input.min), max: Number(input.max), value: Number(input.value) });
+  };
+
+  useEffect(() => {
+    inputRef.current && setTrackPosition(calculateTrackPositionFromInput(inputRef.current));
```
thegrinder, line 103: I wouldn't run this if there is `value` and `onChange ` present

src/components/atoms/InputRange/InputRange.tsx
dfvalero: Why? this is the percentage for the track (from 0 to 1). So, if min is defaulted to 0 and max defaulted to 100, it makes sense to default the trackPosition to 0.5.

src/components/atoms/InputRange/InputRange.tsx
dfvalero: any reason why?

thegrinder: I was thinking about it a bit differently but it makes sense if we rely on reading the value off of the input's ref here 👍

PR #322, opened 2020-05-01
fix(spacing): remove mb prop and document atomic spacing classes

moshie:
Removed the mb prop from `<Heading />`, `<Text />` & `<Badge />`

Update migration method and component documentation.

### DIFF
--- a/etc/docs/migration-v4.md
+++ b/etc/docs/migration-v4.md
@@ -1,0 +1,38 @@
+<h1 align="center"> Migration to 4.0.0 ♻️</h1>
+
+<div align="center"><img src="https://media.giphy.com/media/lpt5YA1gY68yhJjGpm/giphy.gif" align="center" /></div>
+
+# Index
+
+- [Overview](#overview) 🔭
+  - [What has changed](#what-has-changed)
+  - [What has improved](#new-additions)
+  - [What was fixed](#what-was-fixed)
+
+# Overview 🔭
+
+## What has changed?
+
+### 📐 `Spacing`
+
+With V4 introduces a new way of spacing components apart. Using atomic classes there is more granular control over spacing at different breakpoints.
+
+As part of this upgrade the `mb` property on `<Heading />`, `<Text />` & `<Badge />` components has been deprecated.
+
+Before you had to do:
+
+```tsx static
+<Text mb>This is a text element</Text>
+```
+
+Now you can use className to define different margin values
+
+```tsx static
+<Text as="p" className="mb-6">
+  This is a text element
+</Text>
+```
+
+## New additions
+
+## What was fixed
--- a/etc/docs/migration-v4.md
+++ b/etc/docs/migration-v4.md
@@ -17,7 +17,7 @@
 
 With V4 introduces a new way of spacing components apart. Using atomic classes there is more granular control over spacing at different breakpoints.
 
-As part of this upgrade the `mb` property on `<Heading />`, `<Text />` & `<Badge />` components has been deprecated.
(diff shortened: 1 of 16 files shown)


etc/docs/migration-v4.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Overview 🔭
+
+## What has changed?
+
+### 📐 `Spacing`
+
+With V4 introduces a new way of spacing components apart. Using atomic classes there is more granular control over spacing at different breakpoints.
+
+As part of this upgrade the `mb` property on `<Heading />`, `<Text />` & `<Badge />` components has been deprecated.
```
lluia, line 20: `has been removed.` as they won't be able to use it anymore in **v4** 👍🏻

etc/docs/migration-v4.md
```diff
@@ -0,0 +1,40 @@
...
+
+Now you can use className to define different margin values
+
+```tsx static
+<Text as="p" className="mb-6">
+  This is a text element
+</Text>
+```
+
+Note the `as="p"` spans are natuarly inline elements and margin will **not** affect them
```
lluia, line 36: `as="span"` 💅🏻

src/README.md
```diff
@@ -60,6 +60,8 @@ Margin: https://tailwindcss.com/docs/margin/
 
 Padding: https://tailwindcss.com/docs/padding/
 
+The introduction of this new feature means the **mb** property on `<Heading />` & `<Text />` are now deprecated.
```
lluia, line 63: I think we won't need this given we already have the **migration guide** documenting this change ✍🏻 👍🏻

Same for the docs of `<Heading />` and `<Text />`.

src/components/atoms/Text/Text.tsx
```diff
@@ -53,13 +48,6 @@ const Text = styled.span<ITextProps>`
   letter-spacing: 0;
   color: ${({ color = colors.greyDarkest }) => color};
 
-  ${({ mb = false }) =>
-    mb &&
-    `
-    display: block;
-    margin-bottom:   24px
-  `};
```
lluia, line 61: I wonder if we could add `display: block` by parsing `className`:
```js
${({ className, as }) => as === 'span' && className.split(' ').some(clss => clss.includes('mb')) && `
  display: block
`}
``` 
The reasoning being `<Text />` by default renders as a `<span />` and it won't be nice for clients to have to do wrapper components every time:
```js
const BlockText = styled(Text)`display: block`
<BlockText className="mb-6" />
```

src/components/atoms/Text/Text.tsx
```diff
@@ -53,20 +48,20 @@ const Text = styled.span<ITextProps>`
...
   line-height: ${({ size = 'body' }) => lineHeightMap[size]};
   font-weight: ${({ weight = 'regular', capitalize }) => typography.weights[capitalize ? 'bold' : weight]};
 
   font-family: ${typography.primary};
   text-align: ${({ align = 'inherit' }) => align};
   text-transform: ${({ capitalize }) => capitalize && 'uppercase'};
+
+  ${({ className = '', as }) =>
+    (!as || as === 'span') &&
+    className.split(' ').some(clss => clss.includes('mb')) &&
```
lluia, line 61: We probably are gonna need to handle more cases than just `mb`, lets catch-up later 👍🏻

moshie: @lluia 

How do you feel about adding some regex to detect if a spacing class is being used for example: 

```js
/[mp](\:[mp])?[tblr]?-\d+/g
```

PR #337, opened 2020-05-06
feat(inputrange): add controls option to InputRange

dfvalero:
Adding controls to InputRange component.

https://priceless-poincare-18a246.netlify.app/#/Components/Atoms/InputRange

Related JIRA: https://jira.zopa/browse/SHOPDEV-6087

Notes:
- I also fixed the "focus" jump in the Buttons.

### DIFF
--- a/src/components/atoms/Button/Button.tsx
+++ b/src/components/atoms/Button/Button.tsx
@@ -48,7 +48,7 @@ export const buttonStyle = css<IButtonProps>`
   cursor: pointer;
   border-radius: 8px;
   color: ${({ styling = 'primary' }) => colorMap[styling].text};
-  border: none;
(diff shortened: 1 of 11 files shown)


dfvalero:
I really want to add this note here from this article https://www.a11ywithlindsey.com/blog/creating-accessible-range-slider-css

> Unfortunately, using pseudo-elements, we only have support for Edge, IE, and Firefox. There is the pseudo-element ::-ms-fill-lower and ::-moz-range-progress, but none for Chrome. I’ve played with doing this in JavaScript, but it sounds like a performance nightmare and too much math.

I tried to remove the javascript from our slider, but design wants to have this feature in Chrome :disappointed:

PR #338, opened 2020-05-06
Upgrade tooling

poteirard:
- Upgrade react scripts
- Upgrade rollup
- Change browserslist


etc/semantic-release/config.js
```diff
@@ -1,7 +1,12 @@
 module.exports = {
   dryRun: false,
   plugins: [
-    '@semantic-release/commit-analyzer',
+    [
+      '@semantic-release/commit-analyzer',
+      {
+        preset: 'conventionalcommits',
```
poteirard, line 7: by default it's `angular` but we use `conventionalcommits`

package.json
```diff
@@ -37,6 +37,7 @@
     "@types/util-deprecate": "^1.0.0",
     "core-js": "^3.3.5",
     "downshift": "^5.0.0",
+    "jest": "24.9.0",
```
admmasters, line 40: Can we go to 25? They are about to release 26 now..

poteirard: react-scripts force you to have that version

PR #340, opened 2020-05-06
improvement(cardcontent): fix gap at bottom when cards flexed

endymion1818:
Fix for this issue where <Card/> components of a uniform height are required (www.zopa.com & www.zopa.com/savings-accounts) and at some breakpoints the content in one of the cards doesn't wrap, leaving a gap at the bottom:

Screenshot 2020-05-06 at 14 51 50

### DIFF
--- a/src/components/organisms/Card/CardContent/CardContent.tsx
+++ b/src/components/organisms/Card/CardContent/CardContent.tsx
@@ -4,9 +4,6 @@ const CardContent = styled.div`
   background-color: white;
   padding: 16px;
   flex: 1;
-  display: flex;
-  flex-direction: column;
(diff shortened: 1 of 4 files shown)


src/components/organisms/Card/CardContent/CardContent.tsx
```diff
@@ -4,6 +4,9 @@ const CardContent = styled.div`
   background-color: white;
   padding: 16px;
   flex: 1;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
thegrinder, line 9: You can remove this line here and add `flex: 1` to `<Card.Text />`

PR #363, opened 2020-05-07
Beta 13

dfvalero:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@zopauk/react-components",
   "sideEffects": false,
   "description": "Shared react styled components for all the Zopa projects.",
-  "version": "4.0.0--beta.12",
+  "version": "4.0.0-beta.13",
```
moshie, line 5: Should this be `4.0.0--beta.13`?

dfvalero: I don't know, in my note I have this command `npm version --newVersion x.0.0-beta.x`

dfvalero: also, I released 4.0.0-beta.13

poteirard: Actually I don't know why are using this `--beta` thing. I would stick with what `yarn` generates when you run `yarn publish --prerelease`. Which at least it follows a spec

poteirard: @lluia what's the reason of following that beta thing?

dfvalero:
Closed because we released beta.14. A PR will come in a sec

PR #369, opened 2020-05-08
v4 Migration Guide

lluia:
[**Preview**
](https://github.com/zopaUK/react-components/blob/63228b1677e768de7ecb09c3c4bbfae8527e0fc1/etc/docs/migration-v4.md)


etc/docs/migration-v4.md
```diff
@@ -6,16 +6,193 @@
...
 
 # Overview 🔭
 
 ## What has changed?
 
+### 🎨 `Colors`
+
+<img width="983" alt="Screenshot 2020-03-27 at 18 30 26" src="https://user-images.githubusercontent.com/5938217/77783386-0fefde80-7059-11ea-979a-bf3c11f28d27.png">
+
+Colours have changed again. Here's a mapping table hopefully easy your pain through the migration:
```
endymion1818, line 19: 🤣

etc/docs/migration-v4.md
```diff
@@ -6,16 +6,193 @@
...
+<Button loading />
+```
+
+and move away from any custom solution you were using for this.
+
+### 🅰 `Typography`
+
+#### Import
+
+The library doesn't `import` anymore **Open Sans** for you, the `<Fonts />` component has been removed.
```
endymion1818, line 62: suggestion: change to "The library no longer `import`s **Open Sans** for you" ...

etc/docs/migration-v4.md
```diff
@@ -6,16 +6,193 @@
...
+
+```jsx
+import { Heading, Text, Link } from '@zopauk/react-components';
+```
+
+These components retain their API, however `<Heading />`, `<Text />` and `<Link />` have lost their `mb` prop, use [the new atomic classes](#new-additions) instead.
+Since `<Heading />`'s `mb` prop was set to `true` by default, you should add `className="mb-6"` to maintain the existing styling.
+The default for `<Text />` and `<Link />` was `false` so you can just find and replace `mb` prop with `className="mb-6"`.
+
+#### Font sizes
```
endymion1818, line 80: Should we say something about _lead_ here?

thegrinder: `lead` was there before so I'm not sure if it's worth it

endymion1818: 👍

etc/docs/migration-v4.md
```diff
@@ -6,16 +6,199 @@
 
 - [Overview](#overview) 🔭
```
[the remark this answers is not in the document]
thegrinder: fixed

PR #373, opened 2020-05-11
fix(heading): update display size at mobile

moshie:


src/components/atoms/Heading/Heading.tsx
```diff
@@ -67,6 +68,18 @@ const Heading = styled.h1<IStyledHeadingProps>`
   text-align: ${({ align = 'inherit' }) => align};
   font-family: ${typography.primary};
   font-weight: ${({ as }) => typography.weights[['h1', 'display'].includes(as) ? 'extraBold' : 'bold']};
+
+  ${({ as, size }) =>
+    as === 'h1' &&
```
dfvalero, line 73: do we care about the `as === h1`  prop?

moshie: I think so, we don't want all headings turning into h2's at mobile only the top level heading when it's size is set to `display`.

PR #374, opened 2020-05-11
style(footer): correct spacing

dfvalero:


src/components/molecules/ZopaFooter/ZopaFooter.tsx
```diff
@@ -20,7 +20,7 @@ const ZopaFooter = ({ baseUrl = 'https://www.zopa.com', ...rest }: IFooterProps)
   <Footer {...rest}>
```
moshie, line 20: You could achieve the padding on the padding on the footer like this:
```
const ZopaFooter = ({ baseUrl = 'https://www.zopa.com', className = 'pt-10 l:pt-11 mb-9' ...rest }: IFooterProps) => (
  <Footer className={className} {...rest}>
```

That would remove the need for the `Footer` component

dfvalero: nope! we can't do that. If we set a class to `<ZopaFooter />` we will overwrite the internal class.

moshie: Fair enough

PR #377, opened 2020-05-12
Upgrade semantic release

poteirard:
Upgrade semantic release and its plugins

Closes #236

### DIFF
--- a/etc/semantic-release/config.js
+++ b/etc/semantic-release/config.js
@@ -1,24 +1,12 @@
 module.exports = {
   dryRun: false,
+  preset: 'conventionalcommits',
   plugins: [
-    [
-      '@semantic-release/commit-analyzer',
-      {
-        preset: 'conventionalcommits',
-      },
-    ],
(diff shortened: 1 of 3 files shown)


etc/semantic-release/config.js
```diff
@@ -1,24 +1,12 @@
...
+    '@semantic-release/commit-analyzer',
     '@semantic-release/release-notes-generator',
     '@semantic-release/changelog',
     '@semantic-release/npm',
     '@semantic-release/git',
-    [
-      '@semantic-release/github',
-      {
-        assets: [
-          { path: 'dist/index.js', label: 'CJS distribution' },
```
poteirard, line 18: This is old stuff as we compile in different folders now.

package.json
```diff
@@ -103,9 +103,6 @@
     "not ios <11",
     "not dead"
   ],
-  "resolutions": {
-    "semantic-release/@semantic-release/github": "5.5.8"
```
poteirard, line 107: this, in theory, is fixed already

PR #384, opened 2020-05-12
Refactor styleguide + upgrade react-scripts

poteirard:
The latest version of react-scripts wasn't playing very well with react-styleguidist. 
It's not react-styleguidist fault, in fact, I was aware that the library is not responsible for breaking the API of custom styleguides.
 
For this reason, I had to refactor the styleguide. In the process, I also refactored the designs a bit.

Libraries upgraded:
- @testing-library/jest-dom
- react-scripts
- @testing-library/react

Design changes:

Before:
Screen Shot 2020-05-12 at 20 50 44

After:
Screen Shot 2020-05-12 at 21 12 36

Closes: #283


src/styleguide-components/ComponentsList.js
```diff
@@ -1,80 +1,58 @@
...
+
+const HeadingLinkSpan = styled.span`
+  color: ghostwhite;
+  font-weight: bold;
+  font-size: 18px;
+  letter-spacing: -0.5px;
+  margin-top: 12px;
+`;
+
+const List = styled.ul`
```
moshie, line 22: We use this in shopfront as well might be nice as it's own component?

poteirard: This is for the web styleguide only. It's specific for react-styleguidist. More info: https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-styles-of-a-style-guide

PR #398, opened 2020-05-15
fix(styleguide): remove fonts from styleguide

dfvalero:


poteirard:
Thanks! and sorry for my mess!

PR #406, opened 2020-05-18
Fix fonts

poteirard:


styleguide.config.js
```diff
@@ -110,6 +110,14 @@ module.exports = {
   ],
   template: {
     favicon: 'https://s3-eu-west-1.amazonaws.com/cdn.zopa.com/images/201601215/favicon.ico',
+    head: {
+      links: [
+        {
+          rel: 'stylesheet',
+          href: 'https://fonts.googleapis.com/css?family=Open+Sans:400,600,700&display=swap',
```
endymion1818, line 117: I believe we're using weight `800` as well now (display heading)

poteirard: Good catch! Thanks!

PR #415, opened 2020-05-19
fix(fonts): update font stack

moshie:
Update font stack to fonts which have a weight of 800


poteirard:
You also can change it in "styleguide.config.js" line 130

PR #419, opened 2020-05-19
Zds 4 navbar

raff-r:


package.json
```diff
@@ -60,8 +62,8 @@
     "@rollup/plugin-commonjs": "11.1.0",
     "@rollup/plugin-node-resolve": "7.1.3",
     "@rollup/plugin-url": "5.0.0",
-    "@semantic-release/changelog": "5.0.1",
-    "@semantic-release/git": "9.0.0",
```
poteirard, line 64: something is wrong with your merges. Unless you want to downgrade this for some reason

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -1,84 +1,199 @@
...
+      <IconContainer>{withCTA && cta}</IconContainer>
+      {children && <HamburgerMenu open={open}>{children}</HamburgerMenu>}
     </>
   );
 };
 
-export default Navbar;
+const NavbarWrapper: React.FC<INavbarProps> = ({
+  children,
+  overlayLogoWith = <a href={process.env.REACT_APP_ZOPA_LOGO_LINK || 'https://www.zopa.com'} />,
```
poteirard, line 173: I don't think we add env variables in our react components. Do we need it? I would keep the default with 'https://www.zopa.com' only. IF we need to modify this maybe we can use another prop? Same for Navbar action

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -1,84 +1,199 @@
...
-import FlexContainer from '../../../layout/FlexContainer/FlexContainer';
+import { breakpoints } from '../../../../constants/breakpoints';
+import { spacing } from '../../../../constants/spacing';
+import { navbarHeight, mobileNavbarHeight } from '../../../../constants/components';
+import { useViewport } from '../../../../hooks/useViewport';
+import navCurve from '../../../../content/images/nav-curve.svg';
+import Logo from '../../../atoms/Logo/Logo';
+import Icon from '../../../atoms/Icon/Icon';
+import Navbar from '../';
+export interface INavbarProps {
```
admmasters, line 15: Add spacing

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -1,84 +1,199 @@
...
+  /**
```
admmasters, line 16: Also we are meant to be moving away from hungarian notation now, right?

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -1,84 +1,199 @@
...
+const PageNavigation = styled.header`
+  .headroom {
+    top: 0;
+    left: 0;
+    right: 0;
+    z-index: 1;
+    background-color: ${colors.brand};
+    height: ${mobileNavbarHeight}px;
+
+    @media (min-width: ${breakpoints.desktop}px) {
```
poteirard, line 45: Instead of this I think you can use the `minMedia` helper from `helpers/responsiveness.ts`:
```
  ${minMedia.desktop`
  /* your css */
`
```

raff-r:
https://github.com/zopaUK/react-components/pull/422

src/components/organisms/Navbar/Navbar/Navbar.tsx
raff-r: I didn't want to add an additional prop and know we have spoken about having a common list of env variables that we all extend from. Prob got ahead of myself and added it in without discussing with others

src/components/organisms/Navbar/Navbar/Navbar.tsx
raff-r: Did we say what we favour instead?

thegrinder:
Couple of things I noticed while playing with it

I'm aware that it's not easy to make it completely fluid using js but I'm wondering whether we can improve the responsiveness a bit 

navbar_responsive

Also, the fact that we're using `react-headroom` leads to two edge cases that probably need better handling
 
When you open the dropdown and start scrolling, this is what it looks like

Screenshot 2020-05-19 at 17 31 17

src/components/organisms/Navbar/Navbar/Navbar.tsx
poteirard: I will fix this in another PR

poteirard: https://github.com/zopaUK/react-components/pull/428

PR #422, opened 2020-05-19
Navbar Rebrand

raff-r:


raff-r:
https://github.com/zopaUK/react-components/pull/419

thegrinder:
Couple of things I noticed while playing with it

I'm aware that it's not easy to make it completely fluid using js but I'm wondering whether we can improve the responsiveness a bit 

navbar

Screenshot 2020-05-19 at 17 41 06

Also, the fact that we're using `react-headroom` leads to two edge cases that probably need better handling
 
When you open the dropdown and start scrolling, this is what it looks like

Screenshot 2020-05-19 at 17 31 17

Also, I think when a mobile menu is open, we probably shouldn't be hiding the top bar

src/components/organisms/Navbar/Navbar/Navbar.md
```diff
@@ -1,27 +1,373 @@
...
+⚠️ In the examples below the menu will disappear when scrolling down and reappear when scrolling up
 
 ### Examples
 
-- White theme
+- Default theme
 
-```ts { "props": { "style": { "transform": "translate3d(0, 0, 0)", "backgroundColor": "#00B9A7", "border": "2px solid #efefef" } } }
-import { Navbar, Link } from '@zopauk/react-components';
+```ts { "props": { "style": { "transform": "translate3d(0, 0, 0)", "backgroundColor": "#00B9A7", "border": "2px solid #efefef", "overflow": "hidden" } } }
```
thegrinder, line 11: To have an example close to the real life usage, it's probably better if you switch to white background and give it a bit of height as the examples get cut off

src/components/organisms/Navbar/NavbarAction/NavbarAction.tsx
```diff
@@ -0,0 +1,40 @@
...
+  &:hover {
+    color: ${colors.white};
+  }
+`;
+
+const NavbarAction: React.FC<INavbarAction> = ({ ctaText = 'Sign-in' }) => {
+  const { width } = useViewport();
+
+  return (
+    <ButtonLink href={process.env.REACT_APP_ZOPA_MY_ACCOUNT_LINK || 'https://home.zopa.com/'} styling="primary">
```
thegrinder, line 34: I think we might need `onClick` handler here as it might not necessarily be a link

src/components/organisms/Navbar/NavbarLinksList/NavbarLinksList.tsx
```diff
@@ -0,0 +1,86 @@
...
+  ),
+}) => {
+  return (
+    <>
+      {links &&
+        links.map((item: INavigationItem, index: number) => {
+          return !!item.children ? (
+            <span key={`dropdown-${index}`}>{renderDropdown(item, index)}</span>
+          ) : (
+            <span key={`link-${index}`}>{renderLink(item, index)}</span>
```
thegrinder, line 79: I think you can replace span with a `<Fragment />` here cause otherwise you're html nesting gets a bit crazy (i.e. div inside a span)

src/components/organisms/Navbar/NavbarLinksList/NavbarLinksList.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+
+import { INavbarLinkProps } from '../NavbarLink/NavbarLink';
+import Navbar from '../';
+
+export interface INavigationItem {
+  label: string;
+  href?: string;
+  qadata?: string;
+  children?: INavigationItem[];
```
thegrinder, line 10: I have a feeling we might need `onClick` handler here

thegrinder:
There seems to be a problem when resizing the navbar

Screenshot 2020-05-20 at 10 22 04

src/components/organisms/Navbar/Navbar/Navbar.md
```diff
@@ -1,27 +1,374 @@
...
-```ts { "props": { "style": { "transform": "translate3d(0, 0, 0)", "backgroundColor": "#00B9A7", "border": "2px solid #efefef" } } }
-import { Navbar, Link } from '@zopauk/react-components';
+```ts { "props": { "style": { "transform": "translate3d(0, 0, 0)", "backgroundColor": "#FFFFFF", "overflow": "hidden", "height": "500px" } } }
+import { Navbar } from '@zopauk/react-components';
 
-<Navbar left={<Link>left</Link>} center={<Link>center</Link>} right={<Link>right</Link>} />;
+const NAV_ITEMS = [
+  {
+    label: 'Borrow',
+    href: '/loans',
```
thegrinder, line 17: To avoid confusion, it's probably best to remove it the `href` from the dropdown opener. Same thing below

src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
```diff
@@ -1,53 +1,99 @@
...
 }
 
-const StyledNavbarLink = styled(Link)<StyledNavbarLinkProps>`
-  display: inline-flex;
+export interface ChevronContainerProps extends React.HTMLAttributes<HTMLSpanElement> {
+  open: boolean;
+}
+
+export const NavbarLinkStyles = css<StyledNavbarLinkProps>`
   align-items: center;
```
thegrinder, line 31: we used camel case for `linkStyles` and `buttonStyles` - would be nice to be consistent here too

src/components/organisms/Navbar/NavbarLinksList/NavbarLinksList.tsx
```diff
@@ -0,0 +1,37 @@
...
+            key={`dropdown-${index}`}
+            id={`navbar-dropdown-${index}`}
+            label={item.label}
+            items={item.children!}
+            renderItem={({ item, getItemProps }: RenderItemProps) =>
+              renderLink &&
+              renderLink(item, index, {
+                ...getItemProps(),
+                isDropdownLink: true,
+                'data-automation': item.qadata ?? `ZA.${item.qadata}`,
```
thegrinder, line 26: We could probably avoid mapping `qadata` to `data-automation`, or even just spread the props

src/components/organisms/Navbar/NavbarAction/NavbarAction.tsx
```diff
@@ -0,0 +1,51 @@
...
+  `}
+
+  ${minMedia.desktop`
+    ${css`
+      margin-left: ${spacing[2]};
+    `}
+  `}
+`;
+
+const NavbarAction: React.FC<NavbarAction> = ({
```
thegrinder, line 38: no need for `FC` here, you can just do `({ ... }: NavbarAction)`. Having `FC` would technically allow passing `children` to the component

src/components/organisms/Navbar/NavbarAction/NavbarAction.tsx
```diff
@@ -0,0 +1,51 @@
...
+  ctaText = 'Sign in',
+  onClick = () => window.location.assign('https://home.zopa.com/'),
+}) => {
+  const { width } = useViewport();
+
+  return (
+    <StyledButton onClick={onClick} styling={width && width >= breakpoints.desktop ? 'primary' : 'link'}>
+      {width && width >= breakpoints.desktop ? ctaText : <Icon variant={faUser} color={colors.white} size="lg" />}
```
thegrinder, line 46: Actually, you could probably stick to `FC` and pass `children` instead of `ctaText`. I would also wrap the text in a `<span />` and toggle opacity so that when it switches to icon, it still has some sort of label for screen readers

src/components/organisms/Navbar/NavbarAction/NavbarAction.tsx
```diff
@@ -0,0 +1,51 @@
...
+    padding: 0;
+    display: block;
+    width: 100%;
+    height: 100%;
+
+    &:hover:not(:disabled) {
+      background: none;
+    }
+  `}
+
```
thegrinder, line 30: I don't think you actually need any of the styles above

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -1,83 +1,273 @@
...
+                <Icon variant={faBars} color={open ? colors.brand : colors.white} fixedWidth />
+              </HamburgerContainer>
+            ) : (
+              <IconContainer />
+            )}
+            <LogoContainer>
+              <Logo color={colors.brand} height="20px" negative />
+              {overlayLogoWith}
+            </LogoContainer>
+            <IconContainer>{withCTA && cta}</IconContainer>
```
thegrinder, line 260: there is really no need for the `IconContainer` here I believe

src/components/organisms/Navbar/NavbarAction/NavbarAction.md
```diff
@@ -0,0 +1,21 @@
+### Summary
+
+[`<NavbarAction />`](#/Components/Organisms/Navbar/NavbarSignIn) renders the navbar CTA. On Desktop its a [`<Link />`](#/Components/Atoms/Link) styled as a primary button. On Mobile it renders a [user icon](https://fontawesome.com/icons/user?style=solid). You can update the CTA text on desktop
+
+### Examples
+
+- Default
+
+```ts { "props": { "style": { "backgroundColor": "#00B9A7" } } }
```
thegrinder, line 9: can you change the `backgroundColor` to white here? below too

raff-r: I've set it to teal as the icon is white and when on mobile can't see the icon

PR #428, opened 2020-05-20
Enforce naming convention with eslint

poteirard:
- Enforce naming convention with eslint.
- Remove all the "I" and "T" prefixes (hungarian notation)
- I had to change some types/interfaces names Like `Input` to `InputProps`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,13 +17,16 @@ module.exports = {
     // -- TYPESCRIPT
     '@typescript-eslint/camelcase': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
-    '@typescript-eslint/interface-name-prefix': 'off',
+    '@typescript-eslint/naming-convention': [
+      'error',
+      { selector: 'variableLike', format: ['camelCase', 'UPPER_CASE', 'StrictPascalCase'] },
+      { selector: 'typeLike', format: ['StrictPascalCase'] },
+    ],
     '@typescript-eslint/no-explicit-any': 'off', // TODO: Fix all any
     '@typescript-eslint/no-object-literal-type-assertion': 'off',
     '@typescript-eslint/no-empty-interface': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-use-before-define': ['error', { functions: false }],
-    '@typescript-eslint/interface-name-prefix': 'off',
     '@typescript-eslint/no-unused-vars': [
       'error',
       {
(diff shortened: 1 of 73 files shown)


.eslintrc.js
```diff
@@ -17,13 +17,16 @@ module.exports = {
     // -- TYPESCRIPT
     '@typescript-eslint/camelcase': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
-    '@typescript-eslint/interface-name-prefix': 'off',
+    '@typescript-eslint/naming-convention': [
+      'error',
+      { selector: 'variableLike', format: ['camelCase', 'UPPER_CASE', 'StrictPascalCase'] },
+      { selector: 'typeLike', format: ['StrictPascalCase'] },
+    ],
```
poteirard, line 24: This is the important part

src/components/molecules/ExpiryModal/ExpiryModal.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import styled from 'styled-components';
-import Modal, { TModalProps } from '../Modal/Modal';
+import Modal, { ModalProps } from '../Modal/Modal';
```
admmasters, line 3: This is bit stuttery - not for this PR maybe but would be cleaner if we exported these from index.ts in Modal

src/components/atoms/Button/Button.tsx
```diff
@@ -92,23 +92,23 @@ export const buttonStyle = css<IButtonProps>`
...
 `;
 
-const Button: React.FC<IButtonProps> = props => {
+const Button: React.FC<ButtonProps> = props => {
   const { children, loading, styling = 'primary', disabled, ...rest } = props;
   const isLoading = styling !== 'link' && loading;
 
   return (
-    <SButton styling={styling} loading={isLoading} disabled={isLoading || disabled} {...rest}>
+    <ButtonWrapper styling={styling} loading={isLoading} disabled={isLoading || disabled} {...rest}>
```
moshie, line 104: [Question] Why not just `<Wrapper>` since you are in the button file :)

poteirard: I think it's easier to understand the component you are wrapping. But I don't care tbh

src/components/molecules/ExpiryModal/ExpiryModal.tsx
poteirard: You mean having `ModalProps` in src/index.ts`?

PR #431, opened 2020-05-20
fix(footer): fix typo

steedems:
Sorry


steedems:
I have recently moved to vscode and did not realize there was no default spell check (like I had on intellij) - I have now installed Code Spell Checker plugin 😄

PR #442, opened 2020-05-22
Fix error messages

poteirard:
Fix multine error messages

React_components


poteirard:
Moved the icon to top:
React_components

PR #452, opened 2020-06-02
Navbar fixes

raff-r:
Not the most elegant solution to [453](https://github.com/zopaUK/react-components/issues/453) but will avoid breaking changes for now.

Issues were:
- Content wasn't pushed down automatically
- When rendered at 1280px both text and icon wasn't displaying for `<Navbar.Action>`

### DIFF
--- a/src/helpers/responsiveness.ts
+++ b/src/helpers/responsiveness.ts
@@ -16,13 +16,23 @@ const minMedia: Media = {
   desktop: interpolate('desktop', 'min'),
 };
 
-function interpolate(sizeLabel: keyof typeof breakpoints, direction: 'min' | 'max') {
+/* Required due to 1px overlap between minMedia and maxMedia See https://github.com/zopaUK/react-components/issues/453 for details*/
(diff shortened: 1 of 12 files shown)


src/helpers/responsiveness.ts
```diff
@@ -16,13 +16,23 @@ const minMedia: Media = {
   desktop: interpolate('desktop', 'min'),
 };
 
-function interpolate(sizeLabel: keyof typeof breakpoints, direction: 'min' | 'max') {
+/* Required due to 1px overlap between minMedia and maxMedia See https://github.com/zopaUK/react-components/issues/453 for details*/
+const maxEqualToMedia: Media = {
```
raff-r, line 20: It may be a bit extreme but how would people feel about me listing as to be deprecated in next major version? Not sure i've ever introduced a feature that would be deprecated straight away but would prefer people not to get into the habit of using this now and only use it as a last resort to https://github.com/zopaUK/react-components/issues/453

PR #486, opened 2020-06-09
Navbar subheading

raff-r:


src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -276,6 +277,14 @@ const NavbarWrapper: React.FC<NavbarProps> = ({
   const overThreshold = useScrollThreshold(20);
   const [open, setOpen] = useState<boolean>(false);
 
+  useEffect(() => {
+    if (open) {
+      document.body.style.overflow = 'hidden';
+    } else {
+      document.body.style.overflow = 'unset';
+    }
+  }, [open]);
```
thegrinder, line 286: These kinda of things make me think that it would actually be easier to do the mobile navbar with the modal, and probably more accessible too. Probably outside of the scope of this PR though 😅

src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
```diff
@@ -36,36 +36,62 @@ export const navbarLinkStyles = css<StyledNavbarLinkProps>`
...
-    ${({ isDropdownLink }: StyledNavbarLinkProps) =>
+    ${({ isDropdownLink, isDropdownHeading }: StyledNavbarLinkProps) =>
       isDropdownLink &&
+      !isDropdownHeading &&
       css`
-        padding: ${spacing[3]};
+        padding: ${spacing[3]} ${spacing[2]};
+        width: calc(100% - 16px);
+        margin-left: 8px;
+        margin-right: 8px;
```
thegrinder, line 67: I think both spacing variable and hardcoded `px` values are fine here. It would be nice to keep it consistent though - use one approach or the other :)

PR #495, opened 2020-06-11
Navbar critical fixes 

raff-r:


.prettierrc.js
```diff
@@ -2,4 +2,5 @@ module.exports = {
   printWidth: 120,
   singleQuote: true,
   trailingComma: 'all',
+  arrowParens: 'always',
```
poteirard, line 5: This is not needed. Prettier has been upgraded to v2 that's why we now have parentesis everywhere.

From the [docs](https://prettier.io/docs/en/options.html#arrow-function-parentheses):
> First available in v1.9.0, default value changed from avoid to always in v2.0.0 

Try running `yarn install` if you have errors/warns. If you still have problems (as @dfvalero had) try removing `node_modules` and install again.

PR #510, opened 2020-06-16
Add icon to inputs

thegrinder:
Add `startIcon` and `endIcon` to text inputs

image


poteirard:
Shall we force the size of the icon somehow? 
React_components

thegrinder:
hmm @poteirard I did see the past conversation about that but I haven't experienced it on chrome 🤔  I guess one solution would be context and a small wrapper around icon - `<InputIcon />`  - that would grab it from context. I will wook for a way to do it with plain css

tlaak:
I think the difference between the default state and disabled state is too subtle. It's very difficult to see when the field is disabled.

PR #519, opened 2020-06-22
fix(navbar): close hamburger menu on link click

dfvalero:
Design/UX asked to let the user close the hamburger menu when clicking the same link as the current page.

This PR is to do this tricky logic.

### DIFF
--- a/src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
+++ b/src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
@@ -7,6 +7,7 @@ import { minMedia, maxEqualToMedia } from '../../../../helpers/responsiveness';
 
 import Link, { LinkProps } from '../../../atoms/Link/Link';
 import Icon from '../../../atoms/Icon/Icon';
+import { useOpen } from '../OpenProvider';
 
 export interface NavbarLinkStyleProps {
   active?: boolean;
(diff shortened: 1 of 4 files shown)


src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx
```diff
@@ -133,26 +134,37 @@ const NavbarLink: FC<NavbarLinkProps> = React.forwardRef<HTMLAnchorElement, Navb
...
-        <ChevronContainer open={open}>
-          <Icon variant={faChevronDown} color={colors.grey} height="12px" width="12px" />
-        </ChevronContainer>
-      )}
-    </StyledNavbarLink>
-  ),
+  ) => {
+    const { setOpen } = useOpen();
+
+    const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
```
thegrinder, line 144: I think this should go to `<NavbarDropdown/>` into `getItemProps` cause otherwise there is no way to use a custom link

dfvalero: Hi @thegrinder I tested this:

```
const SomeLibraryLink = React.forwardRef((props, ref) => {
  const linkProps = Object.assign({}, props);
  delete linkProps.isDropdownLink;


  return <a {...linkProps} onClick={() => console.log('test')} ref={ref} />;
});
```

and it triggered the console. So, seems that the actual fix is working.

PR #522, opened 2020-06-22
Add warnings for RadioField and SidekickCard

poteirard:
Add deprecation warnings for RadioField and SidekickCard

https://jira.zopa/browse/ZDS-35


src/components/atoms/SidekickCard/SidekickCard.test.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import SidekickCard from './SidekickCard';
 
-describe('<SidekickCard />', () => {
+xdescribe('<SidekickCard />', () => {
```
tlaak, line 6: Why is this excluded?

PR #524, opened 2020-06-22
Redesign Help component

poteirard:
Desktop:

React_components

Desktop - https://www.sketch.com/s/f792aa60-01d9-472d-b53c-c74dd501ef9c/a/GvKm4Y

React_components

Mobile - https://www.sketch.com/s/f792aa60-01d9-472d-b53c-c74dd501ef9c/a/VpJnPo

https://jira.zopa/browse/ZDS-38


src/components/molecules/Help/Help.tsx
```diff
@@ -1,72 +1,87 @@
...
+        <FlexCol xs={12} l={6} className="pb-6">
+          <Card styling="primary">
+            <Card.Content>
+              <FlexRow>
+                <FlexColCenter xs={3} align="center">
+                  <Icon variant={faPhone} color={colors.brand} size="3x" />
+                </FlexColCenter>
+                <FlexCol xs={9}>
+                  <Card.Heading>020 7580 6060</Card.Heading>
+                  <Card.Text>Monday to Thursday (8am to 8pm), Friday (8am to 5pm)</Card.Text>
```
raff-r, line 55: Was there an instance during covid where our call center hours changed? If so i'd suggest we add a prop for timings which has these values as defaults that way if it happens again we wont be required to update the component library to make the change

poteirard: Mmm... Do you think it makes sense? I mean you won't have to upgrade the library but you still will have to manually add the time in every application

poteirard: Also the covid situation is something exceptional, right? I mean we haven't changed that in the past

raff-r: I guess we can release an old version as a patch fix in these scenarios

PR #536, opened 2020-07-02
fix(spacing): fix padding right and add margin auto

raff-r:


src/components/styles/Spacing.ts
```diff
@@ -1,11 +1,14 @@
...
+import { spacing as spacingSizes, grid } from '../../constants';
 import { GridBreakpoints } from '../../constants/grid';
+import { Spacing } from '../../constants/spacing';
 
 type SpacingTypes = 'margin' | 'padding';
 
-const createTopLevelSizes = (type: SpacingTypes) =>
-  Object.keys(sizes).reduce(
+const createTopLevelSizes = (type: SpacingTypes) => {
+  const sizes: Spacing = type === 'margin' ? { ...spacingSizes, auto: 'auto' } : { ...spacingSizes };
```
thegrinder, line 9: no need to spread it like that `{ ...spacingSizes }`. `spacingSizes` will do

src/components/styles/Spacing.ts
```diff
@@ -34,9 +37,12 @@ const createTopLevelSizes = (type: SpacingTypes) =>
     }),
     {},
   );
+};
+
+const createResponsiveSizes = (type: SpacingTypes) => {
+  const sizes: Spacing = type === 'margin' ? { ...spacingSizes, auto: 'auto' } : { ...spacingSizes };
```
thegrinder, line 43: same thing here

PR #564, opened 2020-07-21
Fix spinner on mobile view

katebeavis:
- Fixes spinner on mobile view
- Removes loader on radio field
- Allows a react node to be passed in for an item label on radio field

image

### DIFF
--- a/src/components/molecules/RadioGroupField/RadioGroupField.tsx
+++ b/src/components/molecules/RadioGroupField/RadioGroupField.tsx
@@ -14,10 +14,9 @@ const RadioWrapper = styled.div`
 
 type RadioGroupFieldItem = {
   value: string;
-  label?: string;
+  label?: string | React.ReactNode;
   defaultChecked?: boolean;
   disabled?: boolean;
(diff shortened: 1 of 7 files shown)


src/components/molecules/RadioGroupField/RadioGroupField.tsx
```diff
@@ -14,10 +14,9 @@ const RadioWrapper = styled.div`
 
 type RadioGroupFieldItem = {
   value: string;
-  label?: string;
+  label?: string | React.ReactNode;
```
poteirard, line 17: `React.ReactNode` already accepts `string`

katebeavis: 🤦‍♀️ Should I open a pr with a fix or is it not worth it for such a small thing and leave it to be fixed in another pr?

poteirard: It's not worth it. I commented it just to let you know :) It can be fixed in another PR

PR #570, opened 2020-08-05
Form Section component

raff-r:
A component to group a set of inputs in a form. Will handle consistent spacing between all inputs and can add a title and/or subtitle


src/components/organisms/Form/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC, Children } from 'react';
+import Heading from '../../../atoms/Heading/Heading';
+import Text from '../../../atoms/Text/Text';
+
+type FormSectionProps = {
+  title: string;
```
JuliaHempel, line 6: I would like these to be optional. We don't use them in investments so it would prevent us from using it.

PR #574, opened 2020-08-17
feat(text): adds semibold to the public api for Text

admmasters:
To support then design system more - adding semi bold to the font weights allowed by the Text component.


src/components/atoms/Text/Text.tsx
```diff
@@ -9,7 +9,7 @@ export interface TextProps extends HTMLAttributes<HTMLSpanElement> {
    * The weight of the rendered text.
    * @default 'regular'
    */
-  weight?: 'regular' | 'bold';
+  weight?: 'regular' | 'bold' | 'semiBold';
```
JuliaHempel, line 12: Can you add this to the examples as well?

PR #593, opened 2020-09-24
Card line item

katebeavis:
Adds the `action` prop to `<Card />` and `<CardLineItem />` that can be used together to create an 'action' card
image


src/components/organisms/Card/CardLineItem/CardLineItem.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+
+const CardLineItem = styled.div`
+  padding: 16px;
+`;
```
raff-r, line 5:
```suggestion
const CardLineItem = styled.div.attrs({ className: 'p-4' };
```

steedems (approved):
Awesome thanks! 💯

PR #597, opened 2020-09-28
Update number text

raff-r:
Add an additional font size
Add ability to control title font size
Add text align prop
Add semiBold prop


JuliaHempel (approved):
🍬

PR #599, opened 2020-09-29
fix[numberText]: fix alignment in number text

JuliaHempel:
Make it possible to have left aligned text that is vertically centered in the NumberText component.

### DIFF
--- a/src/components/molecules/NumberText/NumberText.tsx
+++ b/src/components/molecules/NumberText/NumberText.tsx
@@ -1,5 +1,5 @@
 import React, { HTMLAttributes } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/components/molecules/NumberText/NumberText.tsx
```diff
@@ -30,7 +30,7 @@ interface StyleProps {
    * Where the rendered text should be aligned to.
    * @default 'center'
    */
-  align?: 'left' | 'right' | 'center';
+  align?: 'left' | 'right' | 'center' | 'flex-start' | 'flex-end';
```
raff-r, line 33:
```suggestion
  align?: 'left' | 'right' | 'center';
```
Flex-start and flex-end isn't needed as an option if left and right is mapped to these

raff-r (approved):

PR #601, opened 2020-10-02
Zds 41 add container to template

JuliaHempel:
Wrap the content of the template in a container to set the width of the content and center it.

### DIFF
--- a/src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
+++ b/src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
@@ -1,5 +1,4 @@
 import React, { ReactNode } from 'react';
-import styled from 'styled-components';
 
 import { ProgressProps } from '../../../molecules/Progress/Progress';
 import { ProductTemplateTitle } from '../ProductTemplateTitle/ProductTemplateTitle';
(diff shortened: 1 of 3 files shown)


poteirard (approved):

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
```diff
@@ -1,37 +1,37 @@
...
-      <ProductTemplateTitle title={title} subtitle={subtitle} />
-      <ProductTemplateContent>{children}</ProductTemplateContent>
-    </ProductTemplateContainer>
+    <FlexContainer data-automation="ZA.ProductTemplate" className="mb-10">
+      <FlexRow>
+        <FlexCol>
+          <ProductTemplateHeader prevStep={prevStep} progress={progress} />
+          <ProductTemplateTitle title={title} subtitle={subtitle} />
+          <FlexRow justify="center">
+            <FlexCol m={10} xl={contentWidth}>
```
JuliaHempel, line 27: I'm not sure about hardcoding the width here but I also didn't want to add the option of setting the width on every breakpoint due to that making us less consistent across products but I'm happy to change it.

poteirard: Sounds good to me. Let's try to keep it hardcoded as that's what Lianne suggested to have standarized

PR #609, opened 2020-10-13
feat(spinner): update to use gifs

katebeavis:
Use gif for spinner instead of css animation


package.json
```diff
@@ -42,6 +43,7 @@
     "core-js": "^3.3.5",
     "downshift": "^5.2.7",
     "formik": "^2.1.4",
+    "jest-canvas-mock": "^2.3.0",
```
admmasters, line 46: this should presumably be in dev deps

JuliaHempel (approved):
🍰

PR #616, opened 2020-10-15
fix(button): use disabled prop instead of css attribute

steedems:
## The problem 🍿 

I have noticed that the `disabled` attribute is not picked up when using custom anchor element styled with `${buttonStyle}`.

Screenshot 2020-10-15 at 18 01 18

## Suggestion 🧑‍🏭 

Use component `disabled` prop itself to style disabled buttons.

Screenshot 2020-10-15 at 18 28 21


gbkr (dismissed):

raff-r:
This would update the visuals but it wouldn't prevent the link from being clicked. If we are happy to still expect the consumer to prevent that then I guess there is no reason why this can't get merged

steedems:
> This would update the visuals but it wouldn't prevent the link from being clicked. If we are happy to still expect the consumer to prevent that then I guess there is no reason why this can't get merged

yeah I see the point - I think `buttonStyle` has been exported to allow the consumer to style _any_ element like a button, not just an anchor element 🤔 so I guess preventing the click event itself from firing still needs to stay up to the consumer.
If we wanted to support anchor element within the Button we could implement a prop on it and use it like `<Button as="a" ...>` but it would probably still not cover all scenarios like Gatsby, NextJS etc

raff-r:
@steedems I wonder if there is much value in adding this. I feel teams will avoid using a Button link if they want to disable it. We tend to use a button using react-router-dom `history.push` function on an `onClick`.

Where do you plan to use this feature?

steedems:
@raff-r I am using it on a component like this one

```tsx
import { buttonStyle } from "@zopauk/react-components";
import { ButtonProps } from "@zopauk/react-components/types/components/atoms/Button/Button";
import Link, { LinkProps } from "next/link";
import { FC, MouseEvent } from "react";
import styled from "styled-components";

export const StyledA = styled.a`
  ${buttonStyle}
`;

export const ButtonLink: FC<LinkProps & ButtonProps> = ({
  href,
  styling,
  fullWidth,
  children,
  disabled,
}) => (
  <Link href={href}>
    <StyledA
      styling={styling}
      fullWidth={fullWidth}
      disabled={disabled}
      onClick={(e: MouseEvent) => disabled && e.preventDefault()}
    >
      {children}
    </StyledA>
  </Link>
);
```

I could implement my own button with the nextjs equivalent of `history.push` but I'd need to also implement the prefetch logic that nextjs Link includes out of the box - it'd have been handy to avoid doing that but it's definitely not a big deal.
I can close this and implement that way if we prefer 👍

steedems:
Closing - implemented on NextJS as follows

```tsx
import { usePrefetch } from "@utils/hooks/usePrefetch";
import { Button } from "@zopauk/react-components";
import { ButtonProps } from "@zopauk/react-components/types/components/atoms/Button/Button";
import { useRouter } from "next/router";
import React, { FC } from "react";

type ButtonLinkProps = { href: string; replace?: boolean } & ButtonProps;

export const ButtonLink: FC<ButtonLinkProps> = ({
  href,
  replace,
  ...buttonProps
}) => {
  const router = useRouter();
  usePrefetch([href]);
  const handleClick = () => {
    replace ? router.replace(href) : router.push(href);
  };
  return <Button onClick={handleClick} {...buttonProps} />;
};
```

raff-r:
@steedems lets re-open this to allow the styling on buttonLinks

raff-r (dismissed):

src/components/atoms/Button/Button.tsx
```diff
@@ -62,32 +62,36 @@ export const buttonStyle = css<ButtonProps>`
...
 
-  &:disabled {
-    cursor: not-allowed;
-    ${({ loading }) => {
-      if (!loading) {
-        return css`
-          background: ${colors.greyLightest};
-          color: ${colors.grey};
-        `;
+  ${({ disabled, styling = 'primary' }) => {
```
raff-r, line 70: I'd suggest adding this example to the docs https://github.com/zopaUK/react-components/pull/616#issuecomment-710039789

raff-r (approved):

PR #617, opened 2020-10-19
feat(text): as includes figcaption

endymion1818:
A component I'm working on requires a `<figcaption>` element within a `<figure>` element, which contains an image and a string of text. This small PR add the appropriate type to the `<Text/>` component


moshie (approved):
LGTM

PR #619, opened 2020-10-19
fix(product-template): minor design fixes

raff-r:


src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
```diff
@@ -24,11 +24,11 @@ function ProductTemplate({ title, subtitle, children, prevStep, progress, conten
   });
   return (
     <FlexContainer data-automation="ZA.ProductTemplate" className={containerClassnames} gutter={0}>
-      <FlexRow>
+      <FlexRow gutter={0}>
         <FlexCol>
-          <ProductTemplateHeader prevStep={prevStep} progress={progress} />
+          {(!!prevStep || !!progress) && <ProductTemplateHeader prevStep={prevStep} progress={progress} />}
```
admmasters, line 29: Nit: would consider using a ternary here instead of the "&&" syntax - https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx some caveats explained.

JuliaHempel (commented):
🍿

JuliaHempel (dismissed):

admmasters (approved):
🍍

PR #624, opened 2020-10-27
Error pages

raff-r:


moshie (dismissed):

src/components/atoms/Icon/Icon.tsx
```diff
@@ -1,11 +1,34 @@
...
+import { spacing } from '../../../constants/spacing';
+
 interface IconProps extends Omit<FontAwesomeIconProps, 'icon' | 'border'> {
   variant: IconDefinition;
+  bgColor?: Colors[keyof Colors];
 }
 
-export default function Icon({ variant, ...rest }: IconProps) {
-  return <FontAwesomeIcon {...rest} icon={variant} />;
+export const RoundBadge = styled.div<{ color: Colors[keyof Colors] }>`
```
admmasters, line 14: Do we want to allow data-automation support for the wrapping item?

raff-r: Adding data-automation will be added to the svg icon

src/components/templates/ErrorPages/FiveHundred/FiveHundred.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { width = 0 } = useViewport();
+
+  return (
+    <>
+      <FlexRow justify="center">
+        <FlexCol xs="auto" className="mb-7">
+          <StyledIcon bgColor={colors.greyLighter} variant={icon} size="3x" color={colors.greyDark} />
+        </FlexCol>
+      </FlexRow>
+      <Heading as="h2" size={width <= breakpoints.desktop ? 'h4' : 'h3'} align="center" className="mb-4">
```
JuliaHempel, line 31: Can this be an `<H1 />` instead if there is no other H1?

src/components/atoms/Icon/Icon.md
```diff
@@ -32,3 +32,14 @@ import { faCoffee } from '@fortawesome/free-solid-svg-icons';
   <Icon variant={faCoffee} rotation={180} />
 </>;
 ```
+
+You can also wrap an icon with in a circle by setting a `bgColor`
```
admmasters, line 36: within

src/components/atoms/Icon/Icon.tsx
```diff
@@ -1,11 +1,34 @@
...
+  background-color: ${({ color }) => color};
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  width: fit-content;
+  padding: ${spacing[6]};
+`;
+
+export default function Icon({ variant, bgColor, className, ...rest }: IconProps) {
```
admmasters, line 24: Hmmm is it ok that the same className prop is passed to the both the badge and the icon - there may be situations where you want to pass different classes to the badge or the icon.

src/components/templates/ErrorPages/Template/Template.tsx
```diff
@@ -0,0 +1,31 @@
...
+import FlexCol from '../../../layout/FlexCol/FlexCol';
+import { ProductTemplateCard } from '../../ProductTemplate/ProductTemplateCard/ProductTemplateCard';
+
+export interface ErrorTemplateProps {
+  icon?: IconDefinition;
+}
+
+const ErrorTemplate: React.FC = ({ children }) => {
+  return (
+    <FlexContainer data-automation="ZA.FiveHundredTemplate" gutter={0}>
```
katebeavis, line 15: Should this be `ZA.FiveHundredTemplate` as it will also appear for the other templates>

src/components/atoms/Icon/Icon.tsx
raff-r: If a bgColor has been set the className will be added to only the badge and if not set will be added to the icon. Reason for this was for our tailwind spacing classes and ensuring the margin was set to badge rather than the icon.

I didn't want to add an additional prop to be able to set a className to both, but happy to if you feel there would be a need to

admmasters: Fair, lets only add it if we really need the functionality

src/components/templates/ErrorPages/Maintenance/Maintenance.md
```diff
@@ -0,0 +1,13 @@
+### Summary
+
+`<ErrorPageTemplate.Maintenance />` renders the default content for an application at Zopa is in maintenance mode
```
admmasters, line 3: Zopa is in maintenance mode -> Zopa in maintenance mode ?

admmasters (approved):

PR #628, opened 2020-10-29
fix(inputs): bug with iphone safari

moshie:
https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone


moshie:
Would I need to do this for Checkboxes, radio and label elements as well?

endymion1818 (approved):

JuliaHempel (approved):
🍦

PR #635, opened 2020-11-17
feat(product template): render ProductTemplate without title

rorydpayne:
Updated `ProductTemplate` designs with optional title.

`title` now an optional property of `<ProductTemplate />`. Rendering without title will overlay the
content slightly higher leaving only 24px of margin.

#### With Title
desktop_with_title

#### Without Title
desktop_without_title

#### With Title
mobile_with_title

#### Without Title
mobile_without_title

### DIFF
--- a/src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.test.tsx
+++ b/src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.test.tsx
@@ -1,11 +1,11 @@
 import React from 'react';
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, screen, render } from '@testing-library/react';
 import ProductTemplate from '..';
 
 describe('<ProductTemplate />', () => {
   it('renders with all props and content', () => {
     const mockOnBackPressed = jest.fn();
-    const { container, getByLabelText } = render(
+    const { container } = render(
       <ProductTemplate
         title="Product title"
         subtitle="Product subtitle"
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ styleguide/*
 /types
 
 coverage
+/.idea/
```
poteirard, line 14: This is specific to your IDE. You can have a global `gitignore` to ignore your specific software. See [docs](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.test.tsx
```diff
@@ -17,4 +17,45 @@ describe('<ProductTemplate />', () => {
...
     expect(container).toMatchSnapshot();
   });
+
+  it('renders without title', () => {
+    const { container } = render(
+      <ProductTemplate prevStep="prevStep" progress={{ currentStep: 2, totalSteps: 4 }} contentWidth={10}>
+        <ProductTemplate.Card>This is the body of the card</ProductTemplate.Card>
+      </ProductTemplate>,
+    );
+    expect(container).toMatchSnapshot();
```
poteirard, line 27: We are moving away from creating snapshots for everything as it slows down the tests. 

If you want to check if it renders without title the correct approach is to get the title and check is not there. 

If you want to check specific style changes you can also you use the assets `jest-styled-components` is providing like [toHaveStyleRule](https://github.com/styled-components/jest-styled-components#tohavestylerule)

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
```diff
@@ -7,16 +7,31 @@ import { ProductTemplateHeader } from '../ProductTemplateHeader/ProductTemplateH
...
+function overlayElementAbove(marginTop: number) {
+  return `
+    margin-top: -${marginTop - 24}px; 
+    @media (min-width: ${grid.breakpoints.m}px) {
+      margin-top: -${marginTop}px; 
+    }
+    `;
+}
+
+const ContentContainer = styled.div<{ hasTitle: boolean }>(({ hasTitle }) =>
```
poteirard, line 31: provably you don't need an additional div here, you can use `style(FlexRow)` to modify the existing FlexRow

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.test.tsx
rorydpayne: Got it. Was just following the apparent convention

JuliaHempel (commented):
This looks good to me but this was raised with the designers some time ago and they were gonna look into adding designs for it. I'm not sure wha happened there.

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
```diff
@@ -37,16 +52,20 @@ function ProductTemplate({
   return (
     <FlexContainer data-automation="ZA.ProductTemplate" className={containerClassnames} gutter={0}>
       <FlexRow gutter={0}>
-        <FlexCol>
-          {showHeader ? (
+        {showHeader ? (
+          <FlexCol>
             <ProductTemplateHeader prevStep={prevStep} progress={progress} onBackPressed={onBackPressed} />
```
raff-r, line 57: `onBackPressed` isn't needed as `prevStep` can be a ReactElement meaning you can create your own component with onClick handler

src/components/templates/ProductTemplate/ProductTemplate/ProductTemplate.tsx
```diff
@@ -7,17 +7,32 @@ import { ProductTemplateHeader } from '../ProductTemplateHeader/ProductTemplateH
...
   contentWidth?: number;
   onBackPressed?: MouseEventHandler<HTMLAnchorElement>;
 }
 
+function overlayElementAbove(marginTop: number) {
+  return `
+    margin-top: -${marginTop - 24}px; 
+    @media (min-width: ${grid.breakpoints.m}px) {
+      margin-top: -${marginTop}px; 
+    }
```
raff-r, line 28:
```suggestion
    ${minMedia.desktop`
    ${css`
       margin-top: -${marginTop}px; 
    `}
  `}
```
see https://github.com/zopaUK/react-components/pull/635/files#diff-4695613f7b9d259bdf7507895860c5e413b13315ce9c6b2b699d591f2377c117L21 for an example

rorydpayne: I need to use the [m grid breakpoint ](https://github.com/zopaUK/react-components/blob/master/src/constants/grid.ts#L14) here in order to coincide with the [spacing styles](https://github.com/zopaUK/react-components/blob/master/src/components/styles/Spacing.ts). This is to adjust for the different padding-top values of the `<ProductTemplateTitleInnerContainer>` [here](https://github.com/zopaUK/react-components/blob/master/src/components/templates/ProductTemplate/ProductTemplateTitle/ProductTemplateTitle.tsx#L48). Would be worth reviewing this if we can implement a single set of breakpoints across the project.

raff-r (approved):

PR #636, opened 2020-11-17
feat(producttemplate): onBackPressed callback

rorydpayne:
Optional `onBackPressed` callback property on back button click

onbackpressed


rorydpayne:
Will address any @codecov discrepancies after merging #635

poteirard:
Closing since this is not needed.

You can pass a component in `prevStep` with an `onClick`

Issue #637, opened 2020-11-17
ProductTemplate back button extensibility improvement


rorydpayne:
Currently, the `ProductTemplate` component allows you to set a `ProductTemplateProps.prevStep?: string | ReactNode` property in order to display a back navigation button at the top of the page. Passing in a `string` generates a default styled `Link` with `href={prevStep}` as per [the example in the docs](https://zrc.netlify.app/#/Components/Templates/Product). Whereas passing in a `ReactNode` completely replaces the link with your own component.

I think this can be improved. In most cases, clients are going to want the default styled link but may well prefer more control by way of a callback function. At the moment, you would have to implement your own styled back button with `onclick` attribute to achieve this which feels wasteful. 

I propose one of the following:

1. Define `prevStep?: string | ReactNode | () => void`  allowing the client to pass in a callback function to be invoked with the default styled link's `onclick` attribute. This is probably the simplest solution and doesn't change the API much beyond diversifying the `prevStep` union type.

2. Define a new `onBackPressed?: (UIEvent) => void` property of `ProductTemplateProps`. This becomes more of an interceptor as the client can still define an href value for `prevStep`. However, there will be an optional extensibility point before that navigation occurs without any routing responsibility. It means you could define something generic like say...
```tsx { "props": { "style": {  "padding": "0" } } }
const onBackPressed = ({ preventDefault }) => {
  if (confirm('Do you want to exit without saving?')) {
    return;
  }
 preventDefault();
};
```
Still not a breaking change but maybe a bit faffy.

3. If there's no appetite for either of the above changes to `ProductTemplate`, then at least we should expose the back button as an extensible component whilst leaving the `ProductTemplate` alone. This way clients don't need to maintain their own styled navigation buttons.
e.g.
```tsx { "props": { "style": {  "padding": "0" } } }
export const BackButton = ({ href, onClick, text = 'Back' }: BackButtonProps) => (
  <StyledLink href={href} onClick={onClick} aria-label={text}>
    <Icon variant={faChevronLeft} color={colors.greyDark} size="xs" className="mr-2" />
    <BackText>{text}</BackText>
  </StyledLink>
);
```

Let me know what you think.

JuliaHempel:
This sounds good to me, I like the first option the best. It will remove a couple of back buttons created in the different project and still let the projects that use nextjs and Gatsby send in their own link.

rorydpayne:
[...]

PR #639, opened 2020-11-23
feat(producttemplate): onBackPressed callback

rorydpayne:
#637 
Optional `onBackPressed` callback property on back button click

onbackpressed

### DIFF
--- a/src/components/templates/ProductTemplate/ProductTemplateNavigation/ProductTemplateNavigation.test.tsx
+++ b/src/components/templates/ProductTemplate/ProductTemplateNavigation/ProductTemplateNavigation.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
 import { ProductTemplateNavigation } from './ProductTemplateNavigation';
 
 describe('<ProductTemplateNavigation />', () => {
@@ -11,5 +11,12 @@ describe('<ProductTemplateNavigation />', () => {
   it('renders with a component prevStep', () => {
     const { container } = render(<ProductTemplateNavigation prevStep={<>A custom component</>} />);
     expect(container).toMatchSnapshot();
+  });
+
+  it('invokes onBackPressed callback', () => {
+    const mockOnBackPressed = jest.fn();
+    const { getByLabelText } = render(<ProductTemplateNavigation onBackPressed={mockOnBackPressed} />);
+    fireEvent.click(getByLabelText('Back'));
+    expect(mockOnBackPressed).toHaveBeenCalled();
   });
 });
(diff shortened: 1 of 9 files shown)


src/components/templates/ProductTemplate/ProductTemplateNavigation/ProductTemplateNavigation.tsx
```diff
@@ -31,16 +31,16 @@ const BackText = styled(Text).attrs({ color: colors.greyDark })`
...
-export function ProductTemplateNavigation({ prevStep }: ProductTemplateNavigationProps) {
+export function ProductTemplateNavigation({ prevStep, onBackPressed }: ProductTemplateNavigationProps) {
   return (
     <Wrapper className="my-4 m:mt-0" data-automation="ZA.ProductTemplateNavigation">
-      {typeof prevStep === 'string' ? (
-        <StyledLink href={prevStep} aria-label="Back">
+      {typeof prevStep === 'object' ? (
+        prevStep
+      ) : (
+        <StyledLink href={prevStep as string | undefined} aria-label="Back" onClick={onBackPressed}>
```
poteirard, line 40: We have to avoid using type casting.

In this case I think it will look better like this:

```ts
{typeof prevStep === 'string' ? (
        <StyledLink href={prevStep} aria-label="Back" onClick={onBackPressed}>
          <Icon variant={faChevronLeft} color={colors.greyDark} size="xs" className="mr-2" />
          <BackText>Back</BackText>
        </StyledLink>
      ) : (
        prevStep
      )}
```

Then you also have to fix the type of `prevStep` because currently, it supports `boolean` and `number` which is wrong. I know it's not your fault and it was already there... but Can you try to make it work like:

```ts
  prevStep?: ReactElement | string;
```
In theory, this is the correct way, but I'm not 100% sure since my memory is not that good :P

src/components/templates/ProductTemplate/ProductTemplateNavigation/ProductTemplateNavigation.test.tsx
```diff
@@ -12,4 +12,11 @@ describe('<ProductTemplateNavigation />', () => {
     const { container } = render(<ProductTemplateNavigation prevStep={<>A custom component</>} />);
     expect(container).toMatchSnapshot();
   });
+
+  it('invokes onBackPressed callback', () => {
+    const mockCB = jest.fn();
```
poteirard, line 17: We try to avoid acronyms. Can you rename it to `mockOnBackPressed`?

poteirard (commented):
Great job on this @rorydpayne 👏👏

Just left some comments to align in the way we are doing things. But great PR overall!

poteirard (approved):

PR #648, opened 2021-01-05
Update RadioGroupField to allow for id and name

TomGroombridge:
Make `Id` more unique to differentiate between multiple Radio groups with the same values.

Add `name` to allow for simple integration with Formik.


src/components/molecules/RadioField/RadioField.tsx
```diff
@@ -138,9 +138,12 @@ const Input = styled.input<InputStatus & GroupingControlsProps>`
     `}
 `;
 
-export interface RadioField extends FieldProps, InputProps, GroupingControlsProps {}
+export interface RadioField extends FieldProps, InputProps, GroupingControlsProps {
```
admmasters, line 141: Apparently the coverage has gone down based on this - we might want to look at whether a test is needed here.

admmasters: https://github.com/zopaUK/react-components/pull/648/checks?check_run_id=1649671866

src/components/molecules/RadioField/RadioField.tsx
```diff
@@ -138,9 +138,12 @@ const Input = styled.input<InputStatus & GroupingControlsProps>`
...
+  groupLabel?: string;
+}
 
 const RadioField = ({
+  groupLabel = '',
```
poteirard, line 146: Maybe I'm missing something but `groupLabel` is the `name` of the radio group right?

src/components/molecules/RadioField/RadioField.tsx
poteirard: I meant it's not required to merged but we should test the new logic 90% of the times

src/components/molecules/RadioField/RadioField.tsx
TomGroombridge: groupLabel allows us to be unique on when creating the id. We expect there to never be 2 labels the same.

admmasters (approved):

PR #653, opened 2021-01-18
A11y fix nav

endymion1818:
🍿  PR to primarily fix #535 
🎟️  Also fixes #652 & #12 

Note: Jest-axe landmarks rule is now disabled. We're going to have to be a bit more careful about how we use semantic HTML  now.

PR #659, opened 2021-01-20
fix(navbar): add wrapping <ul> in mobile nav

endymion1818:
After yesterdays' PR (#653) I noticed that we were still missing some semantic HTML around the mobile nav.

### DIFF
--- a/src/components/organisms/Navbar/Navbar/Navbar.tsx
+++ b/src/components/organisms/Navbar/Navbar/Navbar.tsx
@@ -267,6 +267,14 @@ const SmallDeviceNavbar = styled.div`
   `}
 `;
 
+const NavItemsWrapper = styled.ul`
+  padding-left: 0;
+
+  > li {
+    display: block;
+  }
+`;
+
 export const NavbarLinksListLink = ({ item: { label, ...rest }, index, props }: NavbarLinksListLinkProps) => (
   <NavbarLink key={`navbar-link-${index}`} {...rest} {...props}>
     {label}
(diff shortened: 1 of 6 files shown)


src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -267,6 +267,14 @@ const SmallDeviceNavbar = styled.div`
   `}
 `;
 
+const NavItemsWrapper = styled.ul`
+  padding-left: 0;
+
+  > li {
```
endymion1818, line 273: This component needs to be `inline-block` for the desktop view, targeting via parent is the only way without splitting the two components which are otherwise very similar.

src/components/organisms/Tabs/TabButtons/__snapshots__/TabButtons.test.tsx.snap
```diff
@@ -145,7 +145,7 @@ exports[`<TabButtons /> renders the buttons on big screens 1`] = `
         <button
           aria-controls="pineapple-content"
           aria-expanded="true"
-          class="c2 c3 py-2 sc-cQFLBn c3 py-2"
+          class="c2 c3 py-2 sc-gojNiO c3 py-2"
```
endymion1818, line 148: I'm not sure why these snapshots were out of date 🤔

poteirard: Interesting. If you run the tests again does it generates a new ID? do you have any idea where this class is coming from?

endymion1818: Yeah, I re-installed my node modules and got a different ID... any ideas?

endymion1818: Yeah, I re-installed my node modules and got a different ID... any ideas?

endymion1818: Okay, I just deleted & cloned down the whole repo again and got the same thing. This is odd.

poteirard: Let's leave it for now and see if it happens to more people

poteirard (approved):

PR #666, opened 2021-03-03
fix(footer): update social links

moshie:
open social links in the footer in a new tab


emilfjellstrom (approved):

admmasters (approved):

OlenaKashuba:
Slightly late for this PR but I have a suggestion that all external links would have attribute title=`${Link name} opens in a new tab`. This is very helpful for users using screen readers. W3Org: "It is recommended that when links are opened to a new window, there is advance warning."

PR #668, opened 2021-03-05
feat(design): design updates

moshie:
- [x] Update close icon on the modal
- [x] Update spacing on modal at mobile 
- [x] Margin under the product template should be: Mob: 40px Desk: 56px
- [x] Update spacing in the product heading template when no progress bar is present
- [x] Increase padding on the right of the accordion header so the text isn't too close to the text.


src/components/molecules/Modal/Modal.tsx
```diff
@@ -48,8 +48,9 @@ const Modal: ModalComponent = ({ children, onRequestClose, showCloseButton = tru
     {showCloseButton && (
       <CrossIcon
```
OlenaKashuba, line 49: The current close button isn't accessible because it is not an interactive element. Let's wrap close icon into the button tag and also add it aria-label "Close" or hidden text "Close". This will make sure the element can be selected via keyboard and properly announce it via screen reader

OlenaKashuba (changes requested):

src/constants/typography.ts
```diff
@@ -3,7 +3,7 @@ const typography = {
   sizes: {
     text: {
       lead: '18px',
-      body: '15px',
+      body: '16px',
```
JuliaHempel, line 6: There's a voice in the back of my head saying we should do this as part of a major release due to the size change could do a lot of difference in how things look and it may be risky to upgrade if for instance a a button gets bigger and won't fit on the row it's on. But I'm happy if we can make this change now.

moshie: I think that's fair now I look at it I just saw that @raff-r already has a PR open for the text-size increase maybe he's already done that testing?

moshie: @JuliaHempel I think the text changes are going to take longer to get tested and approved so I've stripped them out of this PR so we can continue with the design QA.

OlenaKashuba (approved):

src/constants/typography.ts
raff-r: @moshie font size changes should be live early next week

poteirard (approved):

PR #675, opened 2021-04-14
fix(navbar): allows us to suppress user info in navbar with hotjar

MillieClare:
Currently HotJar cannot suppress user names in the navbar. This change means we can now add the suppression tag in react-libs to do so and hide all user information in screen recordings.

It's quite clunky, but `aria-label` can only take a string type, we've added a new property `aria-label` which will be set to `undefined` if it receives a label which isn't of type string.

### DIFF
--- a/src/components/organisms/Navbar/Navbar/Navbar.tsx
+++ b/src/components/organisms/Navbar/Navbar/Navbar.tsx
@@ -22,7 +22,7 @@ import NavbarAction from '../NavbarAction/NavbarAction';
 import NavbarLinksList from '../NavbarLinksList/NavbarLinksList';
 
 export interface NavigationItem {
-  label: string;
+  label: string | React.ReactNode;
   href?: string;
   'data-automation'?: string;
   onClick?: (event?: React.MouseEvent<HTMLAnchorElement>) => void;
--- a/src/components/organisms/Navbar/Navbar/Navbar.tsx
+++ b/src/components/organisms/Navbar/Navbar/Navbar.tsx
@@ -22,7 +22,7 @@ import NavbarAction from '../NavbarAction/NavbarAction';
 import NavbarLinksList from '../NavbarLinksList/NavbarLinksList';
 
 export interface NavigationItem {
-  label: string | React.ReactNode;
+  label: React.ReactNode;
   href?: string;
   'data-automation'?: string;
   onClick?: (event?: React.MouseEvent<HTMLAnchorElement>) => void;
--- a/src/components/organisms/Navbar/NavbarDropdown/NavbarDropdown.tsx
+++ b/src/components/organisms/Navbar/NavbarDropdown/NavbarDropdown.tsx
@@ -78,9 +78,10 @@ export interface NavbarDropdownProps extends DefaultNavbarDropdownProps {
   /** unique id */
   id: string;
   /** dropdown label */
-  label: string;
+  label: string | React.ReactNode;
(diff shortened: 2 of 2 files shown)


src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -22,7 +22,7 @@ import NavbarAction from '../NavbarAction/NavbarAction';
 import NavbarLinksList from '../NavbarLinksList/NavbarLinksList';
 
 export interface NavigationItem {
-  label: string;
+  label: string | React.ReactNode;
```
poteirard, line 25: I think `React.ReactNode` has already `string` as an option so provably we can remove `string |`

poteirard: yep
```ts
 type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

MillieClare: Oh that's nice! I'll update now :)

src/components/organisms/Navbar/NavbarDropdown/NavbarDropdown.tsx
```diff
@@ -78,9 +78,10 @@ export interface NavbarDropdownProps extends DefaultNavbarDropdownProps {
   /** unique id */
   id: string;
   /** dropdown label */
-  label: string;
+  label: React.ReactNode;
```
raff-r, line 81:
```suggestion
  label: string | React.ReactNode;
```

src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -22,7 +22,7 @@ import NavbarAction from '../NavbarAction/NavbarAction';
 import NavbarLinksList from '../NavbarLinksList/NavbarLinksList';
 
 export interface NavigationItem {
-  label: string;
+  label: React.ReactNode;
```
raff-r, line 25:
```suggestion
  label: string | React.ReactNode;
```

src/components/organisms/Navbar/NavbarDropdown/NavbarDropdown.tsx
```diff
@@ -148,7 +149,9 @@ export default class NavbarDropdown extends React.Component<NavbarDropdownProps,
           {label}
         </NavbarLink>
         <NavbarDropdownListContainer open={open}>
-          <NavbarDropdownList aria-label={label}>
+          {/* To avoid an undefined aria-label the fallback to label was added when aria-label property is not defined */}
+          {/* TODO remove fallback to label */}
```
admmasters, line 153: I would make a GH issue for this so we can track the progress of fixing the issue for a future dev.

admmasters: Other than that it looks good to me

src/components/organisms/Navbar/NavbarDropdown/NavbarDropdown.tsx
poteirard: Check the message above :P

poteirard (approved):

PR #678, opened 2021-04-20
Bank details

raff-r:
`<BankDetails />` renders the bank detail with option to copy value to clipboard using [copy-to-clipboard](https://www.npmjs.com/package/copy-to-clipboard)

Screenshot 2021-04-20 at 18 09 38


package.json
```diff
@@ -40,6 +40,7 @@
     "@types/react-modal": "^3.10.6",
     "@types/util-deprecate": "^1.0.0",
     "classnames": "^2.2.6",
+    "copy-to-clipboard": "^3.3.1",
```
admmasters, line 43: Out of interest - any reason why not to use a hook that does this already? e.g. https://github.com/danoc/react-use-clipboard - saves a bit of code - haven't checked the size difference but I suspect they will be pretty similar in size

raff-r: No specific reason. Happy to move to this. Looking at the code its exactly same as I had

JuliaHempel (dismissed):
🍕

src/components/molecules/BankDetails/BankDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+    border-bottom-right-radius: 8px;
+    box-shadow: 0px 1px 0px 0px ${colors.greyLight};
+  }
+`;
+
+const BankDetails: React.FC<BankDetailsProps> = ({ children, copyText }) => {
+  const { isCopied, handleCopy } = useCopyToClipboard();
+
+  return (
+    <ItemWrapper className="px-4 py-5">
```
moshie, line 34: Might I suggest we extend the ItemWrapper so it can accept the className & extra div props. This can be handy for adding extra margins or data automation tags

```suggestion
const BankDetails: React.FC<BankDetailsProps> = ({ children, copyText, className = 'px-4 py-5', ...props }) => {
  const { isCopied, handleCopy } = useCopyToClipboard();

  return (
    <ItemWrapper className={className} {...props}>
```

moshie (changes requested):

moshie (approved):

PR #686, opened 2021-05-05
Fix responsiveness

poteirard:


src/constants/breakpoints.ts
```diff
@@ -1,7 +1,8 @@
 export const breakpoints = {
-  desktop: 1280,
-  phone: 600,
-  tablet: 720,
+  largeDesktop: 1200,
+  desktop: 992,
+  phone: 576,
+  tablet: 768,
```
raff-r, line 5:
```suggestion
  tablet: 768,
  phone: 576,
```
Just being picky but can we have them in order 😛

src/constants/grid.ts
```diff
@@ -11,10 +13,10 @@ interface Grid {
...
   breakpoints: {
     xs: 0,
-    l: 992,
-    m: 768,
-    s: 576,
-    xl: 1300,
+    l: breakpoints.desktop,
+    m: breakpoints.tablet,
+    s: breakpoints.phone,
+    xl: breakpoints.largeDesktop,
```
raff-r, line 19:
```suggestion
    xl: breakpoints.largeDesktop,
    l: breakpoints.desktop,
    m: breakpoints.tablet,
    s: breakpoints.phone,
```
sorry OCD to keep things ordered 😛

src/helpers/responsiveness.ts
```diff
@@ -8,19 +8,22 @@ const maxMedia: Media = {
   phone: interpolate('phone', 'max'),
   tablet: interpolate('tablet', 'max'),
   desktop: interpolate('desktop', 'max'),
+  largeDesktop: interpolate('largeDesktop', 'max'),
```
raff-r, line 11: We should start a migration guide so we don't forget to include any specific content

src/constants/breakpoints.ts
```diff
@@ -1,7 +1,8 @@
 export const breakpoints = {
-  desktop: 1280,
-  phone: 600,
-  tablet: 720,
+  largeDesktop: 1200,
```
raff-r, line 2: where did you get 1200 from? Previously we were using 1280 and 1300 for the grid. I think we should be closer to that to avoid a difference in styling on the larger desktops (as much as possible)

poteirard: I got it from here: https://jira.zopa/browse/ZDS-19

OlenaKashuba: should we confirm breakpoints again with designers?

Issue #691, opened 2021-05-18
Navbar - hamburger menu not a button


JuliaHempel:
On small screens, the hamburger menu is a `<span>` meaning you can't tab into it making it hard for people who navigate without a mouse to open the menu.

Issue #679, opened 2021-04-21
Top-level data-automation in nav items is not rendered


emilfjellstrom:
Currently, the top-level data-automation attributes are not being added 

```
const NAV_ITEMS = [
  {
    label: 'Borrow',
    'data-automation': 'Borrow.topBar.Button',
...
```

Instead a default data attr is added `data-automation="ZA.navbar-item"`
https://github.com/zopaUK/react-components/blob/5fc98a3362e41eef67bc8275960d180ae0a2fd30/src/components/organisms/Navbar/NavbarLink/NavbarLink.tsx#L148

PR #692, opened 2021-05-24
Fix navbar a11y

JuliaHempel:
Will fix https://github.com/zopaUK/react-components/issues/691 and https://github.com/zopaUK/react-components/issues/679

This will:
- change the burger menu to be a button so we can tab into it
- only render the content of the burger menu if the menu is open (to prevent users to tab into hidden elements)
- make sure the `data-automation` attribute is added to all menu links, including the top level links


src/components/organisms/Navbar/Navbar/Navbar.tsx
```diff
@@ -223,9 +225,14 @@ const IconContainer = styled.span`
 
 const HamburgerContainer = styled(IconContainer)<HamburgerContainerProps>`
   background-color: ${({ open }) => (open ? colors.white : 'transparent')};
+  border-radius: ${({ open }) => (open ? 0 : '8px')};
+  &:hover:not(:disabled) {
+    background-color: ${({ open }) => (open ? colors.white : 'transparent')};
+    background: ${({ open }) => (open ? colors.white : 'transparent')};
```
moshie, line 231: Choose only one `background-color` or `background`.

moshie (approved):

admmasters (approved):

PR #707, opened 2021-06-23
fix(radiofield): adding capability to center and make text bold

t-phamm:
**Motivation**

- According to the Zopa design system, a radio button without the control circle should have bold text which is center aligned

**What has changed**

- the `hideControl` prop in RadioField component now decides whether the text will be bold and center aligned 
- Updated relevant snapshots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### [4.43.12](https://github.com/zopaUK/react-components/compare/v4.43.11...v4.43.12) (2021-07-08)
+
+
+### Bug Fixes
+
+* **producttemplate.card:** bottom border visible on all cards mobile ([241014d](https://github.com/zopaUK/react-components/commit/241014d6093ac822fbe0c174238c12ab9248ee4c))
+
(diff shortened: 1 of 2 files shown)


src/components/molecules/RadioField/RadioField.md
```diff
@@ -11,6 +11,8 @@
 - `htmlFor` prop on the `<InputLabel />` component.
 - `id` prop on `<InputRadio />` so that is easily to query (.i.e automated tests).
 
+- `hideControl` prop not only hides the control circle but also centers and text the copy bold, can be used as Option input
```
admmasters, line 14:
```suggestion
- `hideControl` prop hides the control circle, applies bold and centralises the text - can be used as an Option input.
```

src/components/molecules/RadioField/RadioField.md
```diff
@@ -21,7 +23,7 @@ import { RadioField } from '@zopauk/react-components';
 <RadioField label="Option" value="option" name="option" />;
 ```
 
-- Hide control
+- Hide control - also used for Option Input
```
admmasters, line 26:
```suggestion
- Hide control - also used for Option Input.
```

admmasters (dismissed):

PR #714, opened 2021-07-01
relax @babel/runtime dependency to avoid pinning

admmasters:
We should stop pinning this - nextjs uses a newer babel runtime and this somehow ends up in the application.

image


poteirard (approved):

poteirard:
it's in dev deps, does it really affect the prod bundle?

poteirard:
Closing as it seems it's already fixed. Reopen if necessary..

PR #715, opened 2021-07-01
fix(producttemplate.card): bottom border visible on all cards mobile

t-phamm:
**Motivation**

according to the new Zopa designs for one of our projects, we need the the bottom border of each card to be visible in mobile
view instead of only the last one

**What has changed**

applied the bottom border for each card item not only the last one


admmasters (approved):
This probably means the other apps using this component (likely a number of them) should aim to upgrade as well

PR #725, opened 2021-07-27
fix(accordionsection): add visibility to content

emilfjellstrom:
## The challenge 🍿

Toggle visibility on content based on if the accordion is open so that you cant tab into any focusable elements

## The solution 💯

By checking the `visible={sectionPropsRest['aria-hidden']}` we can determine if the accordion is open and toggle `visibility: ${({ visible }) => (visible ? 'hidden' : 'visible')};` with `styled-components` 

## Jira 🎟

- [Jira issue](https://jira.zopa/browse/SHOPDEV-8703)


moshie (approved):
LGTM

PR #738, opened 2021-09-29
footer-a11y-issue

calvin-fung-zopa:


src/components/molecules/ZopaFooter/styles/Heading.tsx
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 import ZopaHeading from '../../../atoms/Heading/Heading';
 
-export const Heading = styled(ZopaHeading).attrs({ as: 'h3', size: 'h6' })`
+export const Heading = styled(ZopaHeading).attrs({ as: 'p', size: 'h6' })`
```
emilfjellstrom, line 4: I think this styled component should wrap the `Text` component as the `Heading` should not allow the attribute `as` as `p`

emilfjellstrom (changes requested):

src/components/molecules/ZopaFooter/styles/Heading.tsx
```diff
@@ -1,6 +1,10 @@
...
-import ZopaHeading from '../../../atoms/Heading/Heading';
+import Text from '../../../atoms/Text/Text';
+import { colors } from '../../../../constants';
 
-export const Heading = styled(ZopaHeading).attrs({ as: 'h3', size: 'h6' })`
+export const Heading = styled(Text).attrs({
+  weight: 'bold',
+  color: colors.greyDarkest,
+})`
   text-transform: uppercase;
```
emilfjellstrom, line 9: There's a `capitalize` prop for the `Text` that we can use here instead of the css

emilfjellstrom (dismissed):
LGTM

src/components/molecules/ZopaFooter/styles/Heading.tsx
calvin-fung-zopa: The capitalised prop has a relationship with the size of the text, hence using that prop would mean that the text can only be small.

poteirard (approved):

PR #747, opened 2022-02-03
fix(zopafooter): removing ISA link

Gabss405:
removing ISA link and moving P2P link to about section
Screenshot 2022-02-03 at 13 45 09


moshie (approved):
LGTM

PR #764, opened 2022-05-19
Locally linking zrc with applications

admmasters:
Branch for creating a poc of theming, locally linking


admmasters:
This was meant to informational

PR #767, opened 2022-05-23
Themable

gbkr:
WIP: Idea for theming, includes Matt's Rollup config


src/components/styles/Theme.tsx
```diff
@@ -0,0 +1,281 @@
...
+import { CardStyling } from '../organisms/Card/Card/Card';
+
+interface ButtonTheme {
+  text: string;
+  bg: string;
+  hover: string;
+}
+
+export type AlertLevel = 'info' | 'alert' | 'warning' | 'success' | 'brand';
+interface CardInfo {
```
admmasters, line 15: spacing

src/components/styles/Theme.tsx
```diff
@@ -0,0 +1,281 @@
...
+  headingSize: string;
+  textSize: string;
+  boxShadowStyle: string;
+  borderStyle: string;
+  backgroundStyle: string;
+}
+
+export interface AppTheme {
+  alert: Record<
```
admmasters, line 24: I'd break this up into named interfaces such as:

```
interface AlertTheme = Record<AlertLevel, { icon: string, background: string; ...etc } 
```

be easier to make individual changes then

gbkr: excellent plan

src/components/styles/Theme.tsx
```diff
@@ -0,0 +1,300 @@
...
+    },
+  },
+};
+
+const ThemeContext = React.createContext<AppTheme>(zopaTheme);
+ThemeContext.displayName = 'ThemeContext';
+
+export function useThemeContext() {
+  const context = React.useContext(ThemeContext);
+  if (!context) {
```
admmasters, line 290: Is there a way we can avoid this, really keen not to get every single app to update, there will be quite a bit of pushback for this

gbkr: So in theory the error below will never be hit unless somebody somehow uses the ThemeProvider within their app and passes in an empty theme prop. The ThemeContext has a default of the Zopa theme, so the context should always exist.

admmasters (approved):
Lets roll

calvin-fung-zopa (approved):

PR #769, opened 2022-05-24
feat: allow passing an undefined theme to ThemeProvider

gbkr:
This allows dynamically passing in a theme - if it's undefined it just defaults to the Zopa theme

### DIFF
--- a/src/components/styles/Theme.tsx
+++ b/src/components/styles/Theme.tsx
@@ -285,16 +285,15 @@ export const zopaTheme: AppTheme = {
 const ThemeContext = React.createContext<AppTheme>(zopaTheme);
 ThemeContext.displayName = 'ThemeContext';
 
-export function useThemeContext() {
-  const context = React.useContext(ThemeContext);
-  if (!context) {
-    throw new Error('No theme provided');
-  }
+export const useThemeContext = () => React.useContext(ThemeContext);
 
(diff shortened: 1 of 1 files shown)


calvin-fung-zopa (approved):
LGTM!

src/components/styles/Theme.tsx
```diff
@@ -285,16 +285,15 @@ export const zopaTheme: AppTheme = {
...
+export const useThemeContext = () => React.useContext(ThemeContext);
 
-export const ThemeProvider = ({ children, theme }: { children: React.ReactNode; theme: Partial<AppTheme> }) => {
-  const mergedTheme: AppTheme = { ...zopaTheme, ...theme };
+export const ThemeProvider = ({
+  children,
+  theme,
+}: {
+  children: React.ReactNode;
+  theme: Partial<AppTheme> | undefined;
```
admmasters, line 295:
```suggestion
  theme?: Partial<AppTheme>;
```

what do you think?

admmasters: Yeah I can see that kind of rationale for this api. Lets roll with it and then we can see how it works out in practice. Given we are the main consumers (its a closed api) - we can maybe revisit

admmasters (approved):

PR #770, opened 2022-05-24
Add theming to heading

calvin-fung-zopa:
## BREAKING CHANGE❗ 
- All <Heading/> uses which are wrapped in styled components should now pass 'forwardedAs' in place of the 'as' prop.
- Docs for forwardedAs: https://styled-components.com/docs/api#forwardedas-prop

## The challenge 🍿

Add theming to Heading component

## The solution 💯

Actual JL component:

Screenshot 2022-05-24 at 18 06 23

Miro designs:

Screenshot 2022-05-24 at 18 45 10

Rezza2 with JL font:

Screenshot 2022-05-25 at 17 04 32

### DIFF
--- a/src/components/atoms/Heading/Heading.tsx
+++ b/src/components/atoms/Heading/Heading.tsx
@@ -27,6 +27,7 @@ export interface OptionalHeadingProps extends HTMLAttributes<HTMLHeadingElement>
 interface HeadingPropsWithAs extends OptionalHeadingProps {
   /**
    * The HTML5 tag you want to render your heading, it's used to determine the size of the heading as well.
+   * @default 'required if 'forwardedAs' is not provided.'
    */
   as: HeadingTags;
 }
@@ -34,7 +35,8 @@ interface HeadingPropsWithAs extends OptionalHeadingProps {
 interface HeadingPropsWithForwardedAs extends OptionalHeadingProps {
   /**
    * The HTML5 tag you want to render your heading, it's used to determine the size of the heading as well.
-   * to be used if Component has been wrapped in styled component in place of as.
+   * To be used if Component has been wrapped in styled components.
+   * @default 'required if 'as' is not provided.'
    */
   forwardedAs: HeadingTags;
 }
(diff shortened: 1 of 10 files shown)


admmasters (approved):

src/components/atoms/Heading/Heading.tsx
```diff
@@ -1,77 +1,83 @@
...
+const StyledHeading = styled.h1<HeadingPropsWithAs & AppThemeProps>`
+  ${({ as, size, theme }) => {
+    if (as) {
+      const tag = size || (as === 'span' ? 'h4' : as);
+      return css`
+        font-size: ${theme.typography.heading.sizes[tag]};
+        line-height: ${theme.typography.lineHeight[tag]};
+        letter-spacing: ${theme.typography.letterSpacingMap[tag]};
+      `;
+    }
```
admmasters, line 55: Do we want to maybe log a warning or something if for whatever reason `as` isn't passed?

calvin-fung-zopa: Would it not be caught by type checking?

admmasters: Then is the `if (as)` redundant?

admmasters: TypeScript thinks the `as` is nullable

OlenaKashuba (commented):
As it is a breaking change, maybe worth adding an example to playground of how the usage of component with forwardedAs

@OlenaKashuba sorry I had to edit the message, it wasn't letting me comment on this. I think your suggestion is good but in this specific case since it's a styled components requirementI think it would be covered in the docs (how styled components should be used with forwardAs: https://styled-components.com/docs/api#forwardedas-prop. Also, I added the props description to the Headings ZRC docs. Do let me know what you think though :)

src/components/atoms/Heading/Heading.tsx
calvin-fung-zopa: Actually you are right, if a user directly passes forwardAs to <Heading/> as will be forwarded to an element which doesn't exist and as wouldn't exist on Heading itself. I guess a warning would make sense :)

Gabss405 (approved):
LGTM

poteirard (approved):

PR #784, opened 2022-06-06
Webpack 5

admmasters:
Bump webpack5, babel and inline create-react-app types, while removing some reliance on create-react-app for bundling as it doesn't work amazingly well with styleguidist.

We retain the `react-scripts test` command as there are some rules that need to be migrated.


moshie (approved):
LGTM 👌

PR #792, opened 2022-06-08
Add theme to CheckBox

OlenaKashuba:
Add theme to CheckBoxField

Add new properties to theme.input.checkBox

### DIFF
--- a/src/components/styles/Theme.tsx
+++ b/src/components/styles/Theme.tsx
@@ -96,13 +96,15 @@ interface InputTheme {
   iconBackgroundColor: string;
   searchInput: {
     boxShadow: string;
-    borderRadiusInput: string;
-    borderRadiusOptions: string;
+    borderRadius: string;
     borderColor: string;
     customIcon: boolean;
-    optionHover: {
-      color: string;
-      backgroundColor: string;
+    options: {
+      borderRadius: string;
+      hover: {
+        color: string;
+        backgroundColor: string;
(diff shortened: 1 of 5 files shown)


src/components/styles/Theme.tsx
```diff
@@ -332,15 +343,26 @@ export const zopaTheme: AppTheme = {
...
-      borderRadiusInput: `8px 8px 0 0`,
-      borderRadiusOptions: `0 0 8px 8px`,
+      borderRadius: `8px 8px 0 0`,
       borderColor: colors.brand,
       customIcon: false,
-      optionHover: {
-        color: colors.white,
-        backgroundColor: colors.brand,
+      options: {
+        borderRadius: `0 0 8px 8px`,
```
admmasters, line 350: Are these 8s meaningful in our scale anywhere, wondering if there are some values we can pull from rather than hardcoding the '8px' in there? If not then I'm happy for this to merge

OlenaKashuba: potentially we could tie it up to `spacing` e.g. `spacing.2` but as it is hardcoded across multiple components, maybe it makes sense to have a specific task to change it in multiple places so it is consistent?

admmasters: Yeah understood, ok lets go with this and then clean up later.

admmasters (approved):

PR #793, opened 2022-06-09
Add theming to ScrollableArea

Gabss405:
- refactored to functional component to be able to access theme context
- `scrollBarTrack` background color, border-radius are themeable
- `scrollBarThumb`  background color, border-radius and border are themeable

### DIFF
--- a/src/components/styles/Theme.tsx
+++ b/src/components/styles/Theme.tsx
@@ -134,6 +134,18 @@ interface NavbarTheme {
 interface ProgressBarTheme {
   color: string;
 }
+interface ScrollableAreaTheme {
+  scrollBarThumb: {
+    borderRadius: string;
+    border: string;
+    bgColor: string;
+  };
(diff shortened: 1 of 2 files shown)


src/components/styles/Theme.tsx
```diff
@@ -368,6 +381,17 @@ export const zopaTheme: AppTheme = {
...
+      border: `2px solid ${colors.brand}`,
+      bgColor: `${colors.brand}`,
+    },
+    scrollBarTrack: {
+      background: `${colors.greyLighter}`,
+      borderRadius: '4px',
+    },
+  },
   typography: {
     primary: `"Open Sans", Roboto, Helvetica, Arial, sans-serif`,
```
OlenaKashuba, line 396: do we need to theme font there too?

Gabss405: ScrollableArea is just a container, the font styles are passed with the children I think

PR #798, opened 2022-06-13
fix(zopafooter): theming bug

Gabss405:
- refactor `Footer` wrapper with Tailwind & add classname to footer type to be able to change padding with theme
- refactor to get text color for `LegalBlock` from theme 
- reorganise footer theme type


src/components/molecules/ZopaFooter/__snapshots__/ZopaFooter.test.tsx.snap
```diff
@@ -380,7 +372,7 @@ exports[`<ZopaFooter /> renders the component with no a11y violations 1`] = `
 }
 
 <footer
-  class="c0"
+  class="c0 pb-9 pt-10 l:pt-11"
```
admmasters, line 375: 👍🏽

PR #800, opened 2022-06-15
Uplfe 343/theme card

gbkr:
https://jira.zopa/browse/UPLFE-343


src/components/organisms/Card/Card/Card.tsx
```diff
@@ -105,4 +107,13 @@ const Card = styled.div<CardProps>`
       : null}
 `;
 
+const Card: React.FC<React.HTMLAttributes<HTMLDivElement> & CardProps> = ({ children, ...props }) => {
```
admmasters, line 110: Only thing here I'd say is [avoid React.FC](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/) - however there is a separate task to move away from React.FC which is used pretty extensively in zrc.

admmasters (approved):

PR #801, opened 2022-06-16
Add unbranded negative spinner

calvin-fung-zopa:
Add unbranded theme for spinner negative:

Screenshot 2022-06-16 at 10 12 28


src/components/atoms/Spinner/Spinner.tsx
```diff
@@ -5,6 +5,7 @@ import actionSpinner from './spinner-action.gif';
 import deprecated from 'deprecated-prop-type';
 import { useThemeContext } from '../../styles/Theme';
 import jlSpinner from '../../../content/images/jl-spinner.gif';
```
admmasters, line 7: I guess we shouldn't reference `jl` specifically in the open source lib, maybe we should rename this to _something_

calvin-fung-zopa: shall we call it 'nonBrandedSpinner' ?

calvin-fung-zopa: since its just black and white it would make sense to refer to it as nonBranded

calvin-fung-zopa: Bear in mind we implemented it like this though ` const animationData = theme?.spinner?.spinnerTheme === 'jl'`. Should we change this too?

calvin-fung-zopa: We could do `theme?.spinner?.spinnerTheme === 'unBranded`?

admmasters (approved):
LGTM!

PR #805, opened 2022-06-21
fix(link): add theme to svg hover and active styles

Gabss405:
- svgs as links take up `hover` and `active` color from theme


admmasters (approved):
LGTM

PR #814, opened 2022-07-08
fix(spinner): passing theme directly into customspinner

Gabss405:


calvin-fung-zopa (approved):

src/components/atoms/Spinner/CustomSpinner/CustomSpinner.tsx
```diff
@@ -31,8 +24,9 @@ const SpinnerAnimation = styled.div<CustomSpinnerProps>`
   .circle-inner {
     transform: rotate(45deg);
     border-radius: 50%;
-    border-width: ${({ fillWidth }) => fillWidth}
-    border-color: ${({ color, negativeColor, styling = 'primary' }) => (styling === 'negative' ? negativeColor : color)}
+    border-width: ${({ theme }) => theme.spinner.customSpinner.fillWidth}
+    border-color: ${({ theme, styling = 'primary' }) =>
```
emilfjellstrom, line 28: Adding enum for these values will make it easier to change in the future i.e `styling = Styling.Primary`

gbkr (approved):

src/components/atoms/Spinner/CustomSpinner/CustomSpinner.tsx
Gabss405: That's a good shout, but it was a union type originally which is fit for purpose I thought. 
Also we had a similar use case recently in another repo and @admmasters suggested we don't use enums based on this: https://fettblog.eu/tidy-typescript-avoid-enums/

admmasters (approved):

src/components/atoms/Spinner/CustomSpinner/CustomSpinner.tsx
admmasters: Yeah we should avoid enums generally. Lots of articles on why, but TLDR: "when compiling an enum, the compiler emits extra JavaScript code that never existed in the original TypeScript code. Few TypeScript features work like this, and each adds a confusing complication to the otherwise simple TypeScript compiler model. For these reasons, we recommend avoiding enums and using unions instead" https://www.executeprogram.com/blog/typescript-features-to-avoid#:~:text=When%20compiling%20an%20enum%2C%20the,enums%20and%20using%20unions%20instead.

admmasters: Lots of examples, but generally in TS land enums are considered a miss-step

PR #815, opened 2022-07-08
Remove react fc

### LINKED ISSUE #763: No longer use React.FC
There are a lot of caveats to using [React FC](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/) - however this is starting to affect newer codebases now as the types are stricter, so certain patterns (e.g using React.FC without children) are throwing new type errors.

Consider using interface types that explicitly list solution, but whoever picks this up might want to discuss.

e.g.
```typescript
interface ButtonProps {
  name: string;
  children: React.Node;
}
```

poteirard:
- Remove React.FC in all the code
- Add eslint rule to avoid adding it again
- Refactor all the namespace components by following [this way](https://react-typescript-cheatsheet.netlify.app/docs/advanced/misc_concerns/#namespaced-components) as it creates better readability and avoid the use of `React.FC`
- Upgraded pnpm to latest 
- Add unused imports rule

Fix #763


tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "downlevelIteration": true,
     "noFallthroughCasesInSwitch": true,
     "types": ["node", "jest", "@testing-library/jest-dom"],
-    "incremental": true,
+    "incremental": false,
```
poteirard, line 21: This is set to true as it had false positive with typechecking.

src/components/templates/ProductTemplate/index.tsx
```diff
@@ -3,14 +3,8 @@ import React from 'react';
...
-const ProductTemplate: ProductTemplateStatic & React.FC<ProductTemplateProps> = (props) => (
-  <ProductTemplateComponent {...props} />
+const ProductTemplate = ({ children, ...rest }: ProductTemplateProps) => (
+  <ProductTemplateComponent {...rest}>{children}</ProductTemplateComponent>
 );
 
-ProductTemplate.Card = ProductTemplateCard;
-
-export default ProductTemplate;
+export default Object.assign(ProductTemplate, { Card: ProductTemplateCard });
```
poteirard, line 10: Another interesting piece of code

admmasters (commented):
Great PR, ❤️ this, do you think it would be possible to break this down into smaller chunks for easier review?

poteirard:
> Great PR, ❤️ this, do you think it would be possible to break this down into smaller chunks for easier review?
> 
> I suppose its all pretty coupled, bar the pnpm one - which could belong in a pr of its own.

Yes, the namespace refactor is to avoid FC. Pnpm change is in a single commit and the eslint imports as well

admmasters (dismissed):

moshie (approved):

moshie (approved):

PR #833, opened 2022-09-01
ZUI-77 || Fix(zopafooter): moved heading from span to h4

khumbon:
## Problem
Accessibility Audit Issue ADR_12398-11: Headings not marked up semantically.

## Solution
Replaced span in ZopaFooter component h4 Heading component to increase accessibility.

Prod vs local change:
image


src/components/molecules/ZopaFooter/ZopaFooter.tsx
```diff
@@ -48,7 +48,7 @@ const ZopaFooter = ({
         {theme.footer.showFooterLinks && (
           <FlexRow className="mb-6">
             <FlexCol xs={12} s={6} l={3} className="mb-8 s:mb-9">
-              <Heading className="mb-4 s:mb-6">What we do</Heading>
+              <Heading className="mb-4 s:mb-6 mt-0">What we do</Heading>
```
moshie, line 51: I think adding the `as="h4"` property makes more sense being added here than in its own component makes it clearer and more readable wdyt?

Also I am not sure why we have `text: uppercase` on them might be worth chatting to design about that since we don't really do it anywhere else to my knowledge?

poteirard (approved):

src/components/molecules/ZopaFooter/ZopaFooter.tsx
khumbon: I'll check with design about this but leave the uppercase for now.
I think I was trying to avoid duplication but having as="h4" makes a lot more sense

poteirard (approved):

PR #850, opened 2022-11-30
ZUI-95 || Fix(navbar and tab): removed circular reference

khumbon:


src/index.ts
```diff
@@ -63,9 +63,7 @@ export { default as LoadingSection } from './components/molecules/LoadingSection
 export * from './components/organisms/Form';
 export type { FormButtonProps } from './components/organisms/Form/FormButton/FormButton';
 export * from './components/organisms/Accordion';
-export { default as Navbar, navbarLinkStyles } from './components/organisms/Navbar/';
 export { default as Card } from './components/organisms/Card';
-export * from './components/organisms/Tabs';
```
poteirard, line 68: If you remove from src/index.ts it won’t be exposed to the library

poteirard (changes requested):

poteirard (approved):

PR #855, opened 2022-12-06
Uplfe1031b

charlielizzy:
there was no ability to adjust font size or weight in the inputtext atom so i've added those props in so that for future components we can amend the size and weight as per design needs. the default remains the same so should have no impact on other uses on inputtext, but the option to adjust font size and weight from the default will now be there

### DIFF
--- a/src/components/atoms/InputText/InputText.tsx
+++ b/src/components/atoms/InputText/InputText.tsx
@@ -97,21 +97,19 @@ const InputText = forwardRef<HTMLInputElement, InputProps>(({ startIcon, endIcon
   const theme = useThemeContext();
 
   return (
-    <>
-      <InputWrapper className={className}>
-        {startIcon && (
-          <IconWrapper startIcon theme={theme}>
-            {startIcon}
-          </IconWrapper>
-        )}
-        <Input startIcon={startIcon} endIcon={endIcon} {...rest} ref={ref} theme={theme} />
-        {endIcon && (
-          <IconWrapper startIcon={false} theme={theme}>
-            {endIcon}
-          </IconWrapper>
-        )}
-      </InputWrapper>
-    </>
+    <InputWrapper className={className}>
+      {startIcon ? (
+        <IconWrapper startIcon theme={theme}>
+          {startIcon}
+        </IconWrapper>
+      ) : null}
+      <Input startIcon={startIcon} endIcon={endIcon} {...rest} ref={ref} theme={theme} />
+      {endIcon ? (
+        <IconWrapper startIcon={false} theme={theme}>
+          {endIcon}
+        </IconWrapper>
+      ) : null}
+    </InputWrapper>
   );
 });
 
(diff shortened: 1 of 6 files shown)


src/components/atoms/InputText/InputText.tsx
```diff
@@ -88,19 +97,21 @@ const InputText = forwardRef<HTMLInputElement, InputProps>(({ startIcon, endIcon
...
-      <Input startIcon={startIcon} endIcon={endIcon} {...rest} ref={ref} theme={theme} />
-      {endIcon && (
-        <IconWrapper startIcon={false} theme={theme}>
-          {endIcon}
-        </IconWrapper>
-      )}
-    </InputWrapper>
+    <>
+      <InputWrapper className={className}>
+        {startIcon && (
```
admmasters, line 102: 👀 we should use a ternary here instead

src/components/atoms/InputText/__snapshots__/InputText.test.tsx.snap
```diff
@@ -22,7 +22,6 @@ exports[`<InputText /> renders the component disabled 1`] = `
   -webkit-transition: 0.2s ease-in-out;
   transition: 0.2s ease-in-out;
   font-family: "Open Sans",Roboto,Helvetica,Arial,sans-serif;
-  font-size: 16px;
```
admmasters, line 25: We should default to `font-size: 16px` so we don't break any consumers of the package

charlielizzy: this bit was automatically deleted when i ran -u to udpate the tests - do you want me to change it back?

src/components/types.ts
```diff
@@ -31,6 +31,8 @@ export interface InputProps extends InputStatus, InputHTMLAttributes<HTMLInputEl
    * Slot for an icon on the right
    */
   endIcon?: ReactNode;
+
+  weight?: number | undefined;
```
moshie, line 35: Don't think the undefined is needed here as `?:` indicates this is an optional param

moshie (changes requested):

src/components/atoms/InputText/InputText.tsx
```diff
@@ -46,15 +56,14 @@ const Input = styled.input<InputProps & { theme: AppTheme }>`
   padding: 0 16px;
   padding-left: ${({ startIcon }) => !!startIcon && '60px'};
   padding-right: ${({ endIcon }) => !!endIcon && '60px'};
-  font-size: ${typography.sizes.text.body};
-  font-weight: ${typography.weights.regular};
+  font-size: ${({ theme, size = 'body' }) => theme.typography.text.sizes[size]};
```
khumbon, line 59: Add a default font-size of 16px

src/components/atoms/InputText/__snapshots__/InputText.test.tsx.snap
charlielizzy: in InputText the default is set to 'body' which is equal to 16px, so no change to the default in reality

src/components/atoms/InputText/InputText.tsx
charlielizzy: 'body' is equal to 16px so there is still a default of 16px this way

admmasters (approved):
LGTM 🚀

moshie (approved):

PR #860, opened 2022-12-13
fix: changed inputprops to fontweight and fontsize to avoid overlap with html native

charlielizzy:
added fontWeight and fontSize props to InputProps to avoid naming of weight and size overlapping with existing props

### DIFF
--- a/src/components/atoms/InputText/InputText.tsx
+++ b/src/components/atoms/InputText/InputText.tsx
@@ -56,7 +56,7 @@ const Input = styled.input<InputProps & { theme: AppTheme }>`
   padding: 0 16px;
   padding-left: ${({ startIcon }) => !!startIcon && '60px'};
   padding-right: ${({ endIcon }) => !!endIcon && '60px'};
-  font-size: ${({ theme, size = 'body' }) => theme.typography.text.sizes[size]};
+  font-size: ${({ theme, fontSize = 'body' }) => theme.typography.text.sizes[fontSize]};
   font-weight: ${({ theme, fontWeight = 'regular' }) => theme.typography.weights[fontWeight]};
   color: ${getInputTextColor};
   border: 1px solid ${getBorderColorByStatus};
--- a/src/components/atoms/InputText/InputText.tsx
+++ b/src/components/atoms/InputText/InputText.tsx
@@ -57,7 +57,7 @@ const Input = styled.input<InputProps & { theme: AppTheme }>`
   padding-left: ${({ startIcon }) => !!startIcon && '60px'};
   padding-right: ${({ endIcon }) => !!endIcon && '60px'};
   font-size: ${({ theme, size = 'body' }) => theme.typography.text.sizes[size]};
-  font-weight: ${({ theme, weight = 'regular' }) => theme.typography.weights[weight]};
+  font-weight: ${({ theme, fontWeight = 'regular' }) => theme.typography.weights[fontWeight]};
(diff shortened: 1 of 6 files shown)


src/components/atoms/InputText/InputText.tsx
```diff
@@ -57,7 +57,7 @@ const Input = styled.input<InputProps & { theme: AppTheme }>`
   padding-left: ${({ startIcon }) => !!startIcon && '60px'};
   padding-right: ${({ endIcon }) => !!endIcon && '60px'};
   font-size: ${({ theme, size = 'body' }) => theme.typography.text.sizes[size]};
```
OlenaKashuba, line 59: you need to use {theme, fontSize} - new prop you renamed

poteirard (approved):

src/components/types.ts
```diff
@@ -32,7 +32,8 @@ export interface InputProps extends InputStatus, InputHTMLAttributes<HTMLInputEl
    */
   endIcon?: ReactNode;
 
-  weight?: number;
+  fontWeight?: number;
+  fontSize?: string;
```
poteirard, line 36: Do we have full freedom of these strings and number? Shall we use a set of options instead?

poteirard (changes requested):

src/components/types.ts
```diff
@@ -32,7 +32,8 @@ export interface InputProps extends InputStatus, InputHTMLAttributes<HTMLInputEl
    */
   endIcon?: ReactNode;
 
-  weight?: number;
+  fontWeight?: number;
```
moshie, line 35: Shouldn't this be a string given how it's used here:

```
${({ theme, fontWeight = 'regular' }) => theme.typography.weights[fontWeight]};
```

OlenaKashuba: no, fontWeight is a number https://github.com/zopaUK/react-components/blob/master/src/components/styles/Theme/ThemeProvider.tsx#L241

OlenaKashuba: https://github.com/zopaUK/react-components/blob/master/src/constants/typography.ts#L32

src/components/types.ts
charlielizzy: good shout - i've added in custom types for FontWeights and FontSizes now :)

src/components/types.ts
moshie: InputProps is being used as the interface for this:

https://github.com/zopaUK/react-components/pull/860/files#diff-7fba51a07023554f00291cb6d526da934112691ee58c2ed0e1f37ffe3e8bd60eR50

and the default here: `fontWeight = 'regular'` is a string which makes no sense

charlielizzy: i've made changes to this now :)

src/components/types.ts
```diff
@@ -32,7 +32,9 @@ export interface InputProps extends InputStatus, InputHTMLAttributes<HTMLInputEl
    */
   endIcon?: ReactNode;
 
-  weight?: number;
+  fontWeight?: keyof AppTheme['typography']['weights'];
```
moshie, line 35: The type of `keyof AppTheme['typography']['weights'];` this is an object not a string or a number?

https://github.com/zopaUK/react-components/blob/master/src/components/styles/Theme/ThemeProvider.tsx#L240-L245

moshie (changes requested):

moshie (approved):

PR #872, opened 2023-02-09
Display Atomic Classes

moshie:
This PR adds atomic classes for the css display property see: https://night-tailwindcss.vercel.app/docs/display


src/content/Display/Display.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { HTMLAttributes } from 'react';
```
moshie, line 1: Ignore this it's only needed to build the documentation

poteirard (approved):

PR #874, opened 2023-02-21
ZUI-25 Feature: added a teal balloon icon

MichelleXBai:


admmasters (commented):
Delete that .DS_Store file and might be worth adding to the .gitignore tbh (slightly more controversial but arguably makes more sense)

PR #913, opened 2023-07-19
Malformed aria on Progress and Accordion

OlenaKashuba:
Accordion: remove aria-label from wrapper
Progress: replace aria-label with aria-valuetext and add a label

### DIFF
--- a/src/components/molecules/Progress/Progress.tsx
+++ b/src/components/molecules/Progress/Progress.tsx
@@ -77,6 +77,16 @@ const ProgressPoint = styled.span<ProgressPointProps>`
   transform: translate(-50%, -50%);
 `;
 
+const HiddenText = styled.span`
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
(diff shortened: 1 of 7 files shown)


src/components/molecules/Progress/Progress.tsx
```diff
@@ -106,11 +116,13 @@ const Progress = ({ totalSteps, currentStep, withStep = false, progressColor, ..
 
   return (
     <ProgressWrapper>
+      <HiddenText id="in-progress">In progress</HiddenText>
```
admmasters, line 119: Might want to make this overridable or it'll be quite easy to have multiple ids which are the same in the DOM.

OlenaKashuba: we shouldn't have more than 1 progress bar on the page, no?

poteirard: Sorry to hijack this thread, can someone approve this https://github.com/ZopaPublic/react-components/pull/936/files :D

OlenaKashuba: In order to get out from being a maintainer you need to help us fix the checks :)

khumbon (approved):

PR #965, opened 2024-01-18
Changesets approach

admmasters:
Testing changesets - use as a reference once the bot has been approved, maybe we can do a mini mob on it so we can have a shared understanding of how this approach works.


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@zopauk/react-components",
+  "name": "@adm-public/react-components",
```
admmasters, line 2: Ignore this - its noise, its so I could test the pushing to the npm registry

PR #1003, opened 2024-05-03
Remove default props

moshie:


admmasters (approved):
Looks good to me - lets ship it!

PR #1021, opened 2024-07-07
Theme navigation bar

admmasters:
- **feat: add banner and logoContainer to NavbarTheme**
- **chore: add some new scripts to support local workflow**
- **chore: changeset**
- **feat: more theming to navbar**


src/components/styles/Theme/ThemeProvider.tsx
```diff
@@ -150,7 +150,7 @@ interface LinkTheme {
   disableTargetIcon: boolean;
 }
 
-interface NavbarTheme {
+export interface NavbarTheme {
```
admmasters, line 153: Product apps have been mirroring this, we should rely on the types exported by the theme provider instead going forwards.

src/components/styles/Theme/ThemeProvider.tsx
```diff
@@ -161,7 +161,23 @@ interface NavbarTheme {
     minHeight: string;
     bgColor: string;
   };
+  logoContainer?: {
+    desktop: {
```
admmasters, line 165: Not 100% on the names here - happy for input?

`minMediaDesktop` might be better to be accurate here

src/components/styles/Theme/ThemeProvider.tsx
```diff
@@ -242,6 +258,11 @@ interface TypographyTheme {
     semiBold: number;
     bold: number;
     extraBold: number;
+
+    // Extra weights to align with partner naming conventions
+    // to discuss with the design/dev team
+    medium?: number;
```
admmasters, line 264: @charlielizzy @MichelleXBai we need to talk about this

moshie (dismissed):

moshie (approved):

